PR #38, opened 2015-05-16
Added properties $iteration and $total to Smarty_Variable

AbcAeffchen:
Sometimes, but not always, the notice 
> Undefined property: Smarty_Variable::$iteration

 and 

> Undefined property: Smarty_Variable::$total

are thrown. I cannot tell why this happens only some times, but I found, that this properties are dynamically set.

### DIFF
--- a/libs/sysplugins/smarty_variable.php
+++ b/libs/sysplugins/smarty_variable.php
@@ -27,7 +27,19 @@ class Smarty_Variable
      * @var int
      */
     public $scope = Smarty::SCOPE_LOCAL;
-
+    /**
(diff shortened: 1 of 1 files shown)


uwetews:
The properties are dynamic by design.

It may have occurred on nested {foreach} loops.
It's already fixed in dev-master and will be tagged as v3.1.24  tomorrow.
Remember to use on nested {foreach}  for each loop different name for item.

AbcAeffchen:
Thanks, I will test this tomorrow. But just to be clear: This did not happen to me on a nested `{foreach}` loop.

PR #61, opened 2015-06-21
use __construct 

DQNEO:
PHP7 raises E_DEPRECATED

```
"E_DEPRECATED:Methods with the same name as their class will not be constructors in a future version of PHP"
```


uwetews:
I have tagged 2.6.29 with this change

DQNEO:
thank you so much!!
:+1:

PR #65, opened 2015-06-29
Create function.get_param.php

Roosso:
Example
`<a href="index.php?page={$page['name']}&pg={$page['n']}{get_params prefix='&amp;' exclude='page,pg'}">{$page['n']}</a>`

### DIFF
--- a/libs/plugins/function.get_param.php
+++ b/libs/plugins/function.get_param.php
@@ -1,0 +1,65 @@
+<?php
+/**
+ * Smarty plugin
+ *
+ * @package Smarty
+ * @subpackage PluginsFunction
+ * @author Jambik <> idea and creations
+ *         alex Roosso <http://www.roocms.com> - mod: added flag noentitys
+ */
+
+/**
+ * Smarty {get_params} function plugin
+ *
+ * Type:     function<br>
+ * Name:     get_params<br>
+ * Purpose:  generate get parameters
+ * Input:
+ *         - prefix  = before string
+ *         - suffix  = after string
+ *         - exclude = GET params wich should be excluded
+ *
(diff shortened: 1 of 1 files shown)


libs/plugins/function.get_param.php
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * Type:     function<br>
+ * Name:     get_params<br>
+ * Purpose:  generate get parameters
+ * Input:
+ *         - prefix  = before string
+ *         - suffix  = after string
+ *         - exclude = GET params wich should be excluded
+ *
+ * @author Jambik <> idea and creations
```
Roosso: sorry...double

AnrDaemon:
Somebody tested the "New pull request" button?
This can be closed as unrelated, IMO.

uwetews:
This is more a user specific plugin subject for a user plugins folder

Roosso:
I dont understand...

See:
In the template, you often have to use dynamic links
For example: /catalog.php?Param1=x&param2=y&page=2
I need to designate a link for the third page, but it also needs to list all the parameters. That would not be used each time the construction of $smarty.get.param1 easier to use the plugin, as I showed.

`<a href="index.php?{get_params prefix='&amp;' exclude='page'}&page=1">Prev</a>
<a href="index.php?{get_params prefix='&amp;' exclude='page'}&page=3">Next</a>`

Roosso:
The documentation did not find anything about the user folder with the plugins. Maybe not looking there !?

AnrDaemon:
Read the entire chapter III. And specifically https://www.smarty.net/docs/en/api.add.plugins.dir.tpl

PR #75, opened 2015-07-31
DateTime object can be instance of DateTimeImmutable since PHP5.5

mrsombre:
In this case we should check DateTimeInterface instead of DateTime

### DIFF
--- a/libs/plugins/shared.make_timestamp.php
+++ b/libs/plugins/shared.make_timestamp.php
@@ -21,8 +21,8 @@ function smarty_make_timestamp($string)
     if (empty($string)) {
         // use "now":
         return time();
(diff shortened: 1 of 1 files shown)


uwetews:
shared.make.timestamp.php shall return (int) not (string).
Why did you change getTimestamp() to format('U') ??

getTimestamp() is also supported by DateTimeImmutable and DateTimeInterface

mrsombre:
This method exists only since 5.3, could be more compatible and do the same
31 Июл 2015 г. 22:59 пользователь "uwetews" <notifications@github.com>
написал:

uwetews:
You are right.
I modified your code so that the result is integer.
Patch is in dev-master and will be part of next release.

PR #96, opened 2015-10-13
do not load "Autoloader" from include path

glensc:
loading it from include path (without abs path) can be dangerous, can load some other project's Autoloader.php instead the right one.


XiongWanTai:
Good ideal. loading something from include path (without abs path) will be dangerous. include once, require once is better.

PR #105, opened 2015-11-09
*Change: {fetch} now can actually use the compiler if the user specified

Zolli:
Hy,

today I try to use this function and noticed, that only fetch content and not compile the fetched file content. The documentation on {include} says that is usable as fetch (Fetch template into a variable using assign), but i think is not the expected behaviour of this function. This behaviour expected by {fetch} (I think). 

I slightly modified the {fetch} behaviour, introduced a new parameter (asTemplate) and if this TRUE use the compiler to copile the given template. 

This modification not break the existing API.

### DIFF
--- a/libs/plugins/function.fetch.php
+++ b/libs/plugins/function.fetch.php
@@ -207,7 +207,15 @@ function smarty_function_fetch($params, $template)
             return;
         }
     } else {
-        $content = @file_get_contents($params['file']);
+        $file = $params['file'];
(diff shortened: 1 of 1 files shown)


uwetews:
{fetch} shall read files not rendered template content.
For rendered template content use {include}.
It can be assigned to variables with {include .... assign='foo'}

I thick you mixed up {fetch} with Smarty::fetch()

PR #181, opened 2016-02-11
Capture KnockoutJS comments

vanderlee:
KnockoutJS uses HTML comments like `<!-- ko text: var -->` and `<!-- /ko -->` to control "virtual element" bindings.
This fix makes Smarty capture these particular comments similarly to MSIE conditional comments.


uwetews:
I committed this including a changelog update

PR #185, opened 2016-02-18
Smarty2

yashchoksi:
Pull


glensc:
looks like bogus PR with no explanation what it's trying to achieve. @uwetews kill it.

PR #195, opened 2016-02-24
Availability to batch variables assigning added

svdmitrij:
Usage:

$view = new Smarty();

$data = [
    'variable1' => $value1,
    'variable2' => $value2,
];

$view->assignArray($data);


tinogo:
That's already possible with the "normal" assign() method. Just pass the array as the first argument to the assign method and do not pass the second argument (defaults to null).
So, no need for this pull request in my mind. :)

svdmitrij:
@tinogo Oh! yep (-:

PR #217, opened 2016-04-14
the uid of source should be calculated

lubobill1990:
the uid of source should be calculated when filepath is get from handler


uwetews:
I did commit your change including change log etc.
It's in the master branch

PR #226, opened 2016-05-03
Allow positional parameters in custom functions

### LINKED ISSUE #164: Positional parameters ?
I'm working on a `{trans}` block for Smarty, and I though it may be cool to accept positional parameters:

```
{trans "Jack" "dull boy"}All work and no play makes %s a %s.{/trans}
```

I imagine positional parameter to be given using integer keys in the `$param` parameter, nicely cohabiting with named parameter (strings key in the `$param` parameter).

But if it's not implemented, there's probably a good reason I miss ?

cdp1337:
This patch is what we've been using for some time for functions in Core Plus such as the date function, `{date $model.created}`.
It adds arguments assigned onto the parameter stack in the next index based on the order that they're set in.

This would fix #164


uwetews:
Your modification does break Smarty's error handling  and is causing the continuous integration on Travis CI to fail.

PR #244, opened 2016-06-11
avoid opcache_invalidate to result in ErrorException

vavrecan:
calling opcache_invalidate will result in ErrorException (Zend OPcache API is restricted by "restrict_api") if opcache.restrict_api is not empty and script is used outside of allowed path. I suggest that there is check which makes sure that function is not called unless opcache.restrict_api is empty - for compatibility with shared webhostings where this configuration property might be common.


uwetews:
I committed this as a patch as opcache_invalidate is used also in several other locations.

PR #249, opened 2016-06-19
Reintroduce assign_by_ref function.

tobya:
Reintroduced the assign_by_ref function as an alias of assignByRef.

Don't know why the existing function was removed as it breaks considerable installed codebase.


tobya:
It appears this is in SmartyBC but I still dont' understand why and I still cant make ti work.

PR #252, opened 2016-06-28
Improve Smarty autoloading.

AnrDaemon:
- Privatize Autoloader class properties.
  There MUST NOT be the use case, need or reason to change them,
  not at runtime, and not from outside the class itself.
- Move *Exception classes to SMARTY_DIR.
  Placing them in sysplugins was, IMO, a mistake.
- Remove now-redundant string comparisons.
  The class is either base (registered in the array and located in the SMARTY_DIR),
  or it is a sysplugin named after "smarty_", so there's no need to check for "smarty" twice.


AnrDaemon:
I wanted to avoid moving files, but it seems there's no way around it.
SMARTY_SYSPLUGINS_DIR could be redefined before autoloader is registered.

Issue #269, opened 2016-08-10
Missing file/class Smarty_Internal_Runtime_ValidateCompiled  since 3.1.30 


meesha81:
force_compile = TRUE   is working

force_compile = FALSE   is not working with this error message:

PHP Fatal error:  Uncaught Error: Class 'Smarty_Internal_Runtime_ValidateCompiled' not found in /usr/share/php/Smarty/sysplugins/smarty_internal_extension_handler.php:130

uwetews:
When upgrading to 3.1.30 please delete manually all existing compiled and cached template files.

PR #283, opened 2016-09-02
Don't hide serious issues

machek:
I've hit same problem as https://github.com/smarty-php/smarty/issues/269
and this small modification would help to show what was the problem


uwetews:
The problem in #269 was not caused by a missing compiled template file.
If the compiled template does exists is tested by method populateCompiledFilepath() which is called before process().
So your patch is needless.

PR #287, opened 2016-09-09
- bugfix {foreach} avoid using empty()

DerSauron:
Using empty() an an object will always return false. So the foreachelse
block will never be reached even with an iterator with zero entries.
Using the total count to determine an empty object.


uwetews:
This is now fixed in the master branch. 
Because of execution time optimization I implemented a little bit  different solution

PR #324, opened 2017-01-06
smarty_internal_runtime_foreach::count returns 0 instead of the corre…

fbett:
…ct count of an ArrayIterator object. iterator_count(object) should be used, instead of count(object), but ArrayIterator implements Countable on PHP 5.6.26


fbett:
Already fixed in commit 61b59a9ec3d736b3473cb644e3e24e9dc14bc31a. Issue #332

PR #331, opened 2017-01-31
Added sorting filters by name, that allows to order filters by passin…

mreiche:
I faced a special problem, where FILTER A needs to be called after FILTER B. The solution was quite simple by keysorting all the filters before applying.

Priorizing filters can be done by:
```
$smarty->registerFilter('output',$callback,'2-FILTER-A');
$smarty->registerFilter('output',$callback,'1-FILTER-B');
```

### DIFF
--- a/libs/sysplugins/smarty_internal_runtime_filterhandler.php
+++ b/libs/sysplugins/smarty_internal_runtime_filterhandler.php
@@ -59,6 +59,8 @@ class Smarty_Internal_Runtime_FilterHandler
         }
         // loop over registered filters of specified type
         if (!empty($template->smarty->registered_filters[$type])) {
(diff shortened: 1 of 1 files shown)


alecpl:
Can't you sort them before registering? It's a BC break.

mreiche:
> Can't you sort them before registering?

The filters are registered dynamically when using a sort of custom plugins. Therefore I don't know the time (and the order) when the filter is beeing registered.

>  It's a BC break.

You mean, If someone has registered the filters in order, this change will break the behaviour? Hmmm... your right. Maybe we can achieve this by adding an additional flag for sorting the filters?

```$smarty->setSortFiltersByName()```

PR #338, opened 2017-02-22
changed the handling of output for default modifier to use isset inst…

alabama:
changed the handling of output for default modifier to use isset instead of @ suppression


alabama:
hm, ok looks like it is harder to solve that problem, cause of the nesting of modifier and other stuff...
Anyone else having similiar issues with the default modifier? Maybe i use it wrong?

uwetews:
Please note that isset() would just work on variables but the modifier could be used on any output like from plugins, object methods, functions etc.

Issue #344, opened 2017-03-16
Should not call class_exists in Smarty_Internal_TemplateCompilerBase::getTagCompiler() or Smarty_Internal_TemplateCompilerBase::callTagCompiler()


kivisade:
Calling `class_exists` in `Smarty_Internal_TemplateCompilerBase::getTagCompiler()` (see [smarty_internal_templatecompilerbase.php#L930](https://github.com/smarty-php/smarty/blob/v3.1.31/libs/sysplugins/smarty_internal_templatecompilerbase.php#L930) for [v3.1.31](https://github.com/smarty-php/smarty/releases/tag/v3.1.31); previously called in `Smarty_Internal_TemplateCompilerBase::callTagCompiler()`, see [smarty_internal_templatecompilerbase.php#L856](https://github.com/smarty-php/smarty/blob/v3.1.30/libs/sysplugins/smarty_internal_templatecompilerbase.php#L856) in [v3.1.30](https://github.com/smarty-php/smarty/releases/tag/v3.1.30) for example) causes Smarty to fail for all tags that do not have `Smarty_Internal_Compile_*` classes (e.g. `html_options` tag) if Smarty is used in a project where a third party PSR-0 class loader is in effect (e.g. `Zend_Loader`).

Changing `class_exists($class_name)` to `$this->smarty->loadPlugin($class_name)` fixes the issue.

uwetews:
This must be a problem of your autoloader. It should never try to load a not existing file but return false in such case. Otherwise class_exists() for not existing classes would always create errors.
The pull request #345  is just a workaround and reduces performance.
It Looks like some wrong configuration of the Zend_Loader

uwetews:
I think you might have optimized the composer autoloader with the --classmap-authoritative option.
This does disable class_exists() tests for none existing classes. Smarty does need this option for its flexibillity 
in handling all different kinds of plugins and extensions.
Use the --optimize-autoloader option for composer autoloader optimization.

However this restriction has been removed in the master branch version of 9.8.2017 Smarty version 3.1.32-dev-19 and will later be included in the 3.1.32 release

PR #345, opened 2017-03-16
Refers smarty-php/smarty#344

kivisade:
Should not call class_exists in Smarty_Internal_TemplateCompilerBase::getTagCompiler() or Smarty_Internal_TemplateCompilerBase::callTagCompiler()


uwetews:
This must be a problem of your autoloader. It should never try to load a not existing file but return false in such case. Otherwise class_exists() for not existing classes would always create errors.
This pull request is just a workaround and reduces performance.

AlBundyLoves69:
May be as a further clarification: I had the same issue. The problem was that my `autoloader` quit when it could not find the class file.

I modified it in that way that when the requested class file name contains `Smarty` then the `autoloader` function just returns instead of displaying an error and calling `die()`.

PR #348, opened 2017-03-23
enable negative values for indent_first

werner-freytag:


wisskid:
Do you have a use-case for this?

werner-freytag:
> Do you have a use-case for this?

Yes, a bullet list, where the first line has a negative indentation relative to the block

wisskid:
I see how one might want this, but I think it's more suitable for a custom plugin.

PR #354, opened 2017-04-25
Fix markdown titles

jcodt:
Just fix README markdown titles


zonuexe:
This problem occurred under the influence of GitHub's Markdown parser change.

PR #363, opened 2017-05-11
Use PHP_VERSION constant instead of phpversion()

ruester:
`phpversion()` may often be disabled in production environments, causing warnings to be generated. Instead `PHP_VERSION` will always be set with the same value and thus having the same effect as `phpversion()` when used in `version_compare()`.


uwetews:
This now in the master branch and will later be included in 3.1.32

PR #368, opened 2017-05-24
Treat callables as callables no matter their origin

### LINKED ISSUE #273: Using closures as registered plugins
Currently, `Smarty::registerPlguin` allows you to supply a Closure (since it is a callable) but when using it in a template it does not compile, resulting in the error `Object of class Closure could not be converted to string`.

This is because (for functions, other types of plugins behave similarly), it hits the following code segment:

``` php
if (!is_array($function)) {
    $output = "{$function}({$_params},\$_smarty_tpl)";
} elseif (is_object($function[ 0 ])) {
    $output =
        "\$_smarty_tpl->smarty->registered_plugins[Smarty::PLUGIN_FUNCTION]['{$tag}'][0][0]->{$function[1]}({$_params},\$_smarty_tpl)";
} else {
    $output = "{$function[0]}::{$function[1]}({$_params},\$_smarty_tpl)";
}
```

The first case gets executed, since a Closure is not an array. This could be solved by adding the following case:

``` php
if ($function instanceof Closure) {
    $output = "\$_smarty_tpl->smarty->registered_plugins[Smarty::PLUGIN_FUNCTION]['{$tag}'][0]({$_params},\$_smarty_tpl)";
} elseif (!is_array($function)) {
// And so forth
```

Could this be supported? I could probably write a patch works for it. This is backwards compatible to at least PHP 5.2, as the class does not have to be loaded (or even exist) for the operator to work.

AnrDaemon:
Should fix #273

### DIFF
--- a/libs/sysplugins/smarty_internal_compile_private_modifier.php
+++ b/libs/sysplugins/smarty_internal_compile_private_modifier.php
@@ -50,7 +50,7 @@ class Smarty_Internal_Compile_Private_Modifier extends Smarty_Internal_CompileBa
                         if (isset($compiler->smarty->registered_plugins[ Smarty::PLUGIN_MODIFIER ][ $modifier ])) {
                             if (is_callable($compiler->smarty->registered_plugins[ Smarty::PLUGIN_MODIFIER ][ $modifier ][ 0 ])) {
                                 $output =
-                                    sprintf('call_user_func_array($_smarty_tpl->registered_plugins[ \'%s\' ][ %s ][ 0 ], [ %s ])',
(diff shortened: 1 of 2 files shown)


AnrDaemon:
Cases 4, 5 and 6 will need similar treatment, but I'm unable to setup testing because `composer.json` missing any dev dependencies whatsoever and I can't figure out correct test cases to trigger the problem.

uwetews:
Case 4 and 5 need no othér treatment as there only functions will be called.
Case 6 the default plugin handler need some more thoughts as the default handler code needs to be called at runtime instead at compile time

AnrDaemon:
Thanks for taking over this. I got the test suite running, but it erroring too much and I yet to get to the errors of interest.

PR #372, opened 2017-06-09
Call preg_match with '0' as flags-parameter

tomterl:
hhvm croaks if preg_match is called with `null` as flags-parameter, because it expects a numerical value. Replacing `null` with `0` in the calls to preg_match that violate this expectation allows smarty to be used under hhvm; we have a mixed setup, running under php-fpm as well as under hhvm on the commandline for tools needing the extra performance and generally saner behaviour of hhvm.

````
Fatal error: Uncaught exception 'ErrorException' with message 'preg_match() expects parameter 4 to be integer, null given' in includes/libs/smarty/3.1.31/sysplugins/smarty_internal_templatelexer.php:310
````

I would be pleased if you would consider incorporating this small fix, or educate me if my fix is violating some standard or good-practice I am unaware of.

Many thanks for the great work on smarty -- it is much appreciated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: php
 
+dist: trusty
+
 php:
   - 5.3
   - 5.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: php
 
-dist: trusty
-
 php:
   - 5.3
   - 5.4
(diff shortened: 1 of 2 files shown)


tomterl:
The failing test is that hhvm is no longer supported on the chosen testing platform.

````
HHVM is no longer supported on Ubuntu Precise. Please consider using Trusty with `dist: trusty
````

uwetews:
Solve with commit https://github.com/smarty-php/smarty/commit/5a8d015ef1315abb4a6befe56abfbef6533c063f

PR #377, opened 2017-07-14
fix 'mkdir(): File exists' error on create directory from parallel processes

a-shpota:


uwetews:
Meanwhile libs/sysplugins/smarty_internal_runtime_writefile.php has been rewritten to retry in case of error.

The fix is in the trunk version and will later be included in 3.1.32

Note:
I recommend also to set $smarty->cache_locking = true;
To prevent on heavily loaded servers possible concurrent cache file updates of same page.

PR #379, opened 2017-07-24
Fix writeFile() mkdir() race condition

fixpunkt:
PHP's `mkdir()` is not concurrency-safe when the `$recursive` flag is enabled. When multiple concurrent processes try to create paths within the same directory structure, interleaving execution of `mkdir()` may cause some of those processes to fail if another process creates a subdirectory they were just about to create. Please see  https://bugs.php.net/bug.php?id=35326 for detailed discussion.

This race condition may cause Smarty template compilation to fail randomly when not all templates have been initialized. Lately, we've observed this problem with increasing frequency in https://github.com/shopware/shopware.

@a-shpota has added code that detects this situation and throws a more meaningful message in 6768340a970beb67422d67e9c885aa226ef95a06. Unfortunately, template compilation will still fail even with these changes applied.

This PR adds a concurrency-safe PHP reimplementation of recursive `mkdir()`. Using this implementation, `writeFile()` will not fail anymore in situations that would have triggered the `mkdir()` race condition before.

### DIFF
--- a/libs/sysplugins/smarty_internal_runtime_writefile.php
+++ b/libs/sysplugins/smarty_internal_runtime_writefile.php
@@ -38,8 +38,8 @@ class Smarty_Internal_Runtime_WriteFile
 
         $_dirpath = dirname($_filepath);
 
-       // if subdirs, create dir structure
-        if ($_dirpath !== '.' && !@mkdir($_dirpath, $_dir_perms, true) && !is_dir($_dirpath)) {
+        // if subdirs, create dir structure
+        if ($_dirpath !== '.' && !@self::ensureDirectoryExists($_dirpath, $_dir_perms)) {
             error_reporting($_error_reporting);
             throw new SmartyException("unable to create directory {$_dirpath}");
         }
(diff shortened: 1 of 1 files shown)


uwetews:
This pull request is rejected as it calls mkdir() on each file write for each subfolder if needed or not. 

Meanwhile libs/sysplugins/smarty_internal_runtime_writefile.php has been rewritten to retry in case of error.

The fix is in the trunk version and will later be included in 3.1.32 

Note:
I recommend also to set $smarty->cache_locking = true;
To prevent on heavily loaded servers possible concurrent cache file updates of same page.

fixpunkt:
Good point about the cache locking, I was unaware of this!

fixpunkt:
I've just tried both of your approaches on a system where I was able to reproduce the issue consistently. 

Enabling `$smarty->cache_locking` had no effect on the issue. I verified that it was set correctly by adding the value of the property to the `throws` clause.

I've also tried your workaround for the race condition. The effect of this was that I was getting `unable to create directory ...` Exceptions all the time now. I also tried increasing the maximum number of retries to 10, which caused the application to stop loading at all.

uwetews:
I have added a clearstatcache() just in case.
Are you sure that compile_dir and cache_dir had right permissions?

fixpunkt:
Yes, I'm sure that the permissions are correct. After all, everything works when I use the fix I've provided in this PR.

I've also tested the version of your fix with `clearstatcache()`. This made the error disappear as well. Weirdly enough, I was unable to reproduce the problem by reverting to the pre-fix version after this without removing the entire cache directory. So the stat cache seems to be involved in the original issue as well.

PR #399, opened 2017-10-22
PHP 7.2:

mambax7:
The each() function is deprecated.
Function create_function() is deprecated

Would these changes work?

### DIFF
--- a/libs/Smarty_Compiler.class.php
+++ b/libs/Smarty_Compiler.class.php
@@ -262,12 +262,12 @@ class Smarty_Compiler extends Smarty {
         reset($this->_folded_blocks);
 
         /* replace special blocks by "{php}" */
-        $source_content = preg_replace_callback($search, create_function ('$matches', "return '"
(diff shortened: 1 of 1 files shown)


uwetews:
It should work, however because of the use of a closure the patch is not compatible with PHP 5.2.

I will release during the comming week 2.6.31 which will implement the callback as method. At the same time I will have also the PHPUnit tests ready including PHP 7.2.
So please wait a few days.

nao-pon:
@uwetews ping. how about it?

uwetews:
I have just released 2.6.31 has the deprecatedfunctions replaced

nao-pon:
Thanks! 👍

vuongmn:
> 
> I have just released 2.6.31 has the deprecatedfunctions replaced

@uwetews ,
Please help me check my Q&A in issue #493 ( Release note of v2.6.31 is sooner than PHP7.2.0)

PR #401, opened 2017-10-24
+html_options placeholder attribute

rudiedirkx:
Demo output is an extra

    <option value="">-- Select state --</option>

### DIFF
--- a/demo/templates/index.tpl
+++ b/demo/templates/index.tpl
@@ -80,7 +80,7 @@ This is an example of the html_options function:
 
 <form>
     <select name=states>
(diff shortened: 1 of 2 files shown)


webmasterMeyers:
I think a better solution would be to do this in your php, `array_unshift($options, ['' => '-- Select state --'])` or '$options =  ['' => '-- Select state --'] + $options

This way you can account for the `''` or `null` or `false` values associated with your "placeholder"

rudiedirkx:
That `array_unshift` works like that?? No way... Empty string, false, null are all the same, so all work. No need for custom accounting. I've just copied html_options and added this. Use it everywhere. Works like a charm.

PR #402, opened 2017-10-26
Let `compileAllTemplates` exit with a non-zero status code if max errors is reached

caugner:
This facilitates running `compileAllTemplates` as part of a Continuous Integration job.


uwetews:
Corresponding patch is now in the master branch and will later be included in 3.1.32

PR #412, opened 2017-11-12
support for ioncube to encrypt tpl

iperfex-team:
doc: http://blog.ioncube.com/2014/11/27/encrytping-smarty-templates/

### DIFF
--- a/libs/sysplugins/smarty_internal_resource_file.php
+++ b/libs/sysplugins/smarty_internal_resource_file.php
@@ -154,7 +154,7 @@ class Smarty_Internal_Resource_File extends Smarty_Resource
      *
      * @return string                 template source
      * @throws SmartyException        if source cannot be loaded
-     */
+     *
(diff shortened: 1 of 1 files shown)


uwetews:
Please understand that we can't add this application dependent patch to the standard Smarty distribution. Note that the patch was also not 100% correct.

You could create a  custom resource plugin (added to the plugins folder) like

resource.ioncube.php
```php
class Smarty_Resource_Ioncube extends Smarty_Internal_Resource_File
{
    public function getContent(Smarty_Template_Source $source)
    {
        if ($source->exists) {
               $res = ioncube_read_file($source->filepath);
               if (!is_int($res)) {
                   return $res;
               }
        }
        throw new SmartyException('Unable to read ' . ($source->isConfig ? 'config' : 'template') .
                                  " {$source->type} '{$source->name}'");
     }
}
```
and use
```php
$smarty->default_resource_type = 'ioncube';
```

PR #414, opened 2017-11-21
Improved foreach generators support.

AnrDaemon:


uwetews:
Can you please explain what benefit you see to use a closure instead of assigning arrays?

AnrDaemon:
In my case, closure is a database lazy load iterator that potentially fetch thousands of rows.
Prefetching them is a bit silly, as you could imagine. Using generator lets me avoid scattering $smarty->isCached() checks throughout the code, and overall create a more modular framework.
Of course, this PR is merely a quick hack with limited application (i.e. my iterator allows for an optional initial node specifier, which this hack does not allow), naturally this should be resolved at parsing level.
But even this hack is better than current state, as it allows for repeatable generator invocation.

uwetews:
I like the general idea but I think there are a couple of things to consider befor it can be integrated in the official release. But I will investigate how it could look like.

Have already the first issue
The patch should be first thing in Smarty_Internal_Runtime_Foreach::init() to allow the correct generation of the $smarty.foreach....   variable.

AnrDaemon:
Do you mean the failed `$total` assignment?
I see how it could be a problem, but I have no good idea how to deal with it. Not very familiar with internals, and don't have much time at present.

uwetews:
I will look into it later.

uwetews:
Why did you use a closure?
If you would use a generator it would not require your patch.

See http://php.net/manual/de/language.generators.php

AnrDaemon:
I use anonymous function with a number of parameters passed through use() construct.
I could probably turn it into its own class, but frankly… Too much code, where a function already does the same.

AnrDaemon:
This seems to work better with latest Smarty versions. May be not perfect, but I can live with it.

PR #416, opened 2017-12-05
Use Composer classmap instead of bootstrap.

AnrDaemon:


AnrDaemon:
Custom bootstrap is inferior to classmap or PSR autoload in that it is not available during Composer hooks, for example.

uwetews:
bootstrap was used to install Smarty's autoloader because of some problems which poped up when a user used certain autoloader optimization options. This is still subject for some detailed investigations.
For the time being I keep the current implementation.

AnrDaemon:
I did not propose to remove bootstrap. By all means, keep it supplied for projects with manual Smarty installation.
But for Composer projects, Composer offers a better option.

PR #423, opened 2018-02-08
Using call_user_func_array to allow passing by reference.

ggvanleeuwen:
Using call_user_func it's not possible to pass a variable as reference, while it is being done. This results in warnings being thrown. When switching to call_user_func_array it is possible to pass by reference.

See docs: https://secure.php.net/manual/en/function.call-user-func-array.php and https://secure.php.net/manual/en/function.call-user-func.php

### DIFF
--- a/libs/sysplugins/smarty_internal_runtime_filterhandler.php
+++ b/libs/sysplugins/smarty_internal_runtime_filterhandler.php
@@ -60,7 +60,7 @@ class Smarty_Internal_Runtime_FilterHandler
         // loop over registered filters of specified type
         if (!empty($template->smarty->registered_filters[ $type ])) {
             foreach ($template->smarty->registered_filters[ $type ] as $key => $name) {
-                $content = call_user_func_array($template->smarty->registered_filters[ $type ][ $key ], [$content, &$template]);
+                $content = call_user_func_array($template->smarty->registered_filters[ $type ][ $key ], array($content, &$template));
             }
         }
         // return filtered output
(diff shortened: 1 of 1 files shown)


AnrDaemon:
Can you please review your patch against current base?

ggvanleeuwen:
It seems builds for php7 and up are currently not passing due to a dependency issue.

AnrDaemon:
It is line 28 from .travis.yml.
I'll see what can be done.

AnrDaemon:
I think I have a solution.

AnrDaemon:
I think I'll have to postpone this.
I may need to switch to Linux box for further development, as under Windows, `smarty/smarty-phpunit` looks like total mess, and doesn't run as is without much of the massaging.
Should untangle it first, and see where it could lead after that.
Not right now, sorry. May be in a month. If anybody want to pick up, there's [open pull request](https://github.com/smarty-php/smarty-phpunit/pull/1) that at least fixes version selection.

AnrDaemon:
Mental note:
1. Deprecate smarty-dev package.
2. Order cross-dependencies to fit the bills.
3. Ensure tests can be run from Smarty itself, not requiring any convoluted processing past "composer install".
4. Preferable from "composer test" straight away.

uwetews:
Since PHP 5 there is no real benefit passing objects by reference. Smarty 2 did use it because of PHP 4 which did copy the whole object when passing not by reference. PHP 5 does use pointer.
In Smarty 3 all reference usage for passing Smarty or template objects has been removed in for example Plugins, Filter etc
By mistake the documentation of the registerFilter() shows in the technical note still the reference.

For good programming practice you should pass variables by reference if needed.
You should change your filter not requiring reference.

AnrDaemon:
It's not about objects passed by reference. It's about function arguments passed by reference.

wisskid:
@ggvanleeuwen is this still an issue? If so, could you provide a use case?

ggvanleeuwen:
@wisskid I'm unsure. I'm no longer using this.

PR #424, opened 2018-02-12
Add option to prevent double escaping

mschop:


AnrDaemon:
Then add option to prevent triple escaping, then quadriple…
You are supposed to feed it the raw data. If it is somehow escaped prior to hitting the output stream, this is not Smarty, this is your problem. You screwed somewhere in your code.
Remove bogus escapings from your code and Smarty would just work.

mschop:
@AnrDaemon. Your answer does not make sense to me. What should triple encoding even be in this context? This parameter only prevents `htmlspecialchars` from reencoding already encoded entities. Without this parameter to `htmlspecialchars` this `&#x27;` would become that: `&amp;#x27;`. No problems should arise from adding this option to smarty. It doesn't reduce the security level of smarty and does not add any significant complexity. The problem of preventing double encoding is already solved by PHP core devs for the method `htmlspecialchars`. They saw the requirement for this parameter, so why shouldn't you?

Background information: At the time when creating this pull request, I was improving the security of this application by enabling html escaping by default. The application had escaped values in the database and migrating those values to contain no html entities would be a complex task. Migrating millions of datasets can be problematically for standard software. By providing this option to smarty users, you would simplify improving the security of applications.

AnrDaemon:
Fix your broken application. This is not a Smarty problem.

mschop:
@AnrDaemon I think your argument is an ivory tower argument. Imagine a REST-API from a third party saas solution that you can't control. If such third party application returns html entities in the response, than you would have to first unescape the values before assigning them to smarty for preventing double encoding. This adds superflous complexity to the application and can decrease application performance compared to if smarty would use the prevent double encoding feature from `htmlspecialchars`.

So in my opinion this IS a smarty problem, even if the original source of the problem is caused somewhere else.

AnrDaemon:
How is this in any way related to Smarty? If you get data you know is broken in a certain way, you fix it before passing to Smarty.
And you report it to the data owner that their data is broken so they can either fix it or document it.

PR #438, opened 2018-05-02
Fixes Smarty->compileAllTemplates() which was ignoring its $extension parameter

### LINKED ISSUE #437: [3.1.32] Smarty::compileAllTemplates ignores `$extension` parameter
We run the following in a CI job:

```php
<?php
$smarty = new Smarty();
$smarty->setCompileDir(...);
$smarty->setTemplateDir(...);


$smarty->compileAllTemplates('.tpl', true);
```

This has always worked fine.

However, after upgrading from 3.1.31 to 3.1.32, `compileAllTemplates` reports several errors in files without the `.tpl` extension.

caugner:
Fixes #437


uwetews:
It's now in the master branch

caugner:
@uwetews Any chance to have a 3.1.33 release (with this change) or are there blocking issues that need to be addressed first?

PR #452, opened 2018-06-12
Update PSR-2

dusta:
Automatically correct coding standard violations


uwetews:
The original patch did incorrectly  format switch/case tags 
. PSR-2 coding standards require indention at case

PR #458, opened 2018-07-04
Fixed bug in date_format modifier

peterader:
If the date is NULL it would output the current time instead of simply returning. Further, the use of "|default:"-"}" is no longer possible due to this bug.

### DIFF
--- a/libs/plugins/modifier.date_format.php
+++ b/libs/plugins/modifier.date_format.php
@@ -42,9 +42,9 @@ function smarty_modifier_date_format($string, $format = null, $default_date = ''
         }
         $is_loaded = true;
     }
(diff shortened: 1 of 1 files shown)


juangacovas:
I can confirm this changed some months ago and was unexpected

mangasf:
Hi, 

When will this change be integrated ? 

Thanks.

hngl:
When will 3.1.34 be released? This bug just screwed us over...

wisskid:
@uwetews could you please reelase 3.1.34? We had broken builds and had to revert to 3.1.31 too. Thanks!

XWB:
Ping @uwetews This bug is rather annoying and forces us to update a lot of templates.

PR #461, opened 2018-07-16
separator only between items

webmasterMeyers:
For functions `html_checkboxes` & `html_radios` attribute `$separator`
A seperator should only appear between the items, not just after.

Example of a problem scenario, `separator="</li><li>"` to put each item in a list row, would result in an extra empty row at the end.

If you want the extra separator at the end of your list, put it in the template.


webmasterMeyers:
miss use of rtrim

PR #462, opened 2018-07-16
html_checkboxes & html_radios separator fix

webmasterMeyers:
For functions `html_checkboxes` & `html_radios` attribute `$separator`
A seperator should only appear between the items, not just after.

Example of a problem scenario, `separator="</li><li>"` to put each item in a list row, would result in an extra empty row at the end.

If you want the extra separator at the end of your list, put it in the template.

### DIFF
--- a/libs/plugins/function.html_checkboxes.php
+++ b/libs/plugins/function.html_checkboxes.php
@@ -172,7 +172,7 @@ function smarty_function_html_checkboxes($params, Smarty_Internal_Template $temp
     }
 
     // remove 'separator' from last index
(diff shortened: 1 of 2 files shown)


webmasterMeyers:
This is still not working right, at least I think I am missing the entire last array item.  I guess this is a WIP - I have no idea how to change the pull request other than to close it

webmasterMeyers:
This needs a `!empty($sep...)` check

wisskid:
This is working as the examples in the documentation clearly suggest: https://www.smarty.net/docs/en/language.function.html.radios.tpl so we can't change it.
We should probably fix the Description for the separator argument in the docs, it is appended to each element instead of separating the elements.

Issue #466, opened 2018-07-24
assign - Function has standard value for second parameter


Sironheart:
The assign function is using the following declaration

`function assign($tpl_var, $value = null) { ... }`

I believe that there is no reason why there should be the possibility to not a assign a specified value! also there would be a clear Linting Error at PHPStorm, Sublime, etc if you do not give it a second parameter.

AnrDaemon:
Your wording making it unclear, what are you complaining about.
The documentation clearly states that if first parameter is an array, the second parameter is not used.

PR #469, opened 2018-08-01
#466

Sironheart:
Fixing a problem where no error is thrown, if no values assigned

### DIFF
--- a/libs/sysplugins/smarty_internal_data.php
+++ b/libs/sysplugins/smarty_internal_data.php
@@ -96,7 +96,8 @@ abstract class Smarty_Internal_Data
      */
     public function assign($tpl_var, $value = null, $nocache = false)
     {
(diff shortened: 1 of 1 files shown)


libs/sysplugins/smarty_internal_data.php
```diff
@@ -96,6 +96,10 @@ public function __construct()
      */
     public function assign($tpl_var, $value = null, $nocache = false)
     {
+        if(is_array($tpl_var) && $value !== null) {
```
AnrDaemon: I would change the test to 
```suggestion
        if(is_array($tpl_var) && isset($value)) {
```
While semantically the same, `isset()` conveys the meaning of the test more clearly.

AnrDaemon:
An actual exception text would be more useful.

wisskid:
Closing due to inactivity.

PR #473, opened 2018-08-12
some phpsorm inspections fixes

Gemorroj:


uwetews:
debug.tpl was not correctly patched

Issue #478, opened 2018-08-20
{$smarty.capture.$somevar} no longer works


ymTm7KuhCnIjkZAl2x2m2:
The following code and output should display the issue.  

Template:
```
Smarty: {$smarty.version|escape}
<br />
PHP: {$phpversion}
<br />
{capture name=FooBar assign=FooBar}I am the foobar{/capture} 
{assign txt 'FooBar'} 
1: {$smarty.capture.FooBar}<br />
2: {$smarty.capture.$txt}<br />
3: {$smarty.capture[$txt]}<br />
4: {$txt}<br />
<!-- just in case it's caps lock related --> 
{capture name=baz assign=baz}I am the baz{/capture} 
{assign txt2 'baz'} 
5: {$smarty.capture.baz}<br />
6: {$smarty.capture.$txt2}<br />
7: {$smarty.capture[$txt2]}<br />
8: {$txt2}<br />
```

Output:
```
Smarty: 3.1.27 
PHP: 5.6.36 
1: I am the foobar
2: I am the foobar
3: I am the foobar
4: FooBar
5: I am the baz
6: I am the baz
7: I am the baz
8: baz
```

```
Smarty: 3.1.33-dev-5 
PHP: 5.6.36 
1: I am the foobar
2: 
3: 
4: FooBar
5: I am the baz
6: 
7: 
8: baz
```

uwetews:
Please note that the Smarty special variables like $smarty.foo .... are no real variables and are not necessarily implemented as array. It is not specified that variable capture names can be used like 
```php $smarty.capture.$foo ```
However it did work unintended in some previous versions of Smarty

Access to variable capture blocks could be achieved by

```php
{capture assign='bar'}
    buh
{/capture} 
 ....
{$foo}  // with $foo = 'bar'
```

Anyway I did add the variable capture name now as a feature

PR #481, opened 2018-08-24
Fix for $smarty.capture.$varhere, issue #478

ymTm7KuhCnIjkZAl2x2m2:
Adds a very specific regex to let $smarty.capture$somevar to work once again.  It appears to me by the fact that $parameter[1] is $_smarty_tpl->tpl_vars['bar']->value that this was intended to work.


uwetews:
Please see comments at https://github.com/smarty-php/smarty/issues/478

PR #492, opened 2018-09-27
Allow empty cache_lifetime, cache_id, compile_id for subtemplates

moroz1999:
cache_lifetime, cache_id, compile_id for subtemplates should not only be set, but should be not empty for subtemplate caching to be enabled. 

At the moment even if cache_id is provided as false or null, the subtemplate caching is turned on for the template. This leads to a bloated code like:
{if $cacheId = getSomeDynamicCacheId($someFactor)}
    {include file="my.tpl" cache_id=$cacheId}
{else}
    {include file="my.tpl"}
{/if}

With this change this would be done as:
{include file="my.tpl" cache_id=getSomeDynamicCacheId($someFactor)}

So caching won't be turned on if that getSomeDynamicCacheId would return false or null.

### DIFF
--- a/libs/sysplugins/smarty_internal_compile_include.php
+++ b/libs/sysplugins/smarty_internal_compile_include.php
@@ -158,21 +158,21 @@ class Smarty_Internal_Compile_Include extends Smarty_Internal_CompileBase
         } else {
             $_new_caching = Smarty::CACHING_LIFETIME_CURRENT;
         }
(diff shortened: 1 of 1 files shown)


AnrDaemon:
This PR can not be accepted as it disallows "0" as cache_id for example.

moroz1999:
Understood the idea about why this can't be used, thanks,

PR #494, opened 2018-10-01
FIX: throw $e will swallow exception stacktrace

quamis:
use throw new Exception with 3'rd parameter filled in, to throw a new
exception

### DIFF
--- a/libs/sysplugins/smarty_internal_templatebase.php
+++ b/libs/sysplugins/smarty_internal_templatebase.php
@@ -253,7 +253,7 @@ abstract class Smarty_Internal_TemplateBase extends Smarty_Internal_Data
             if (isset($_smarty_old_error_level)) {
                 error_reporting($_smarty_old_error_level);
             }
-            throw $e;
+            throw new Exception("Error while fetching template", null, $e);
         }
     }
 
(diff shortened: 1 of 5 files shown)


wisskid:
@quamis what problem are you trying to fix?

quamis:
Exceptions thown by non-smarty-code in tpl's had their stacktrace truncated, because smarty was catching exceptions and re-throwing them without attaching the old exception. This method of re-throwing didn't allow propagating the original stacktrace from the original exception, and it made debugging really hard, because all exceptions seemed to occur in the smarty core.

quamis:
Sorry, I've re-reviewd the changes and I wasn't remembering the whole picture.

Smarty was doing a lot of `throw $e`, but this will throw back the original exception, without recording where and why it was changed and re-thrown(and keeping only the original stacktrace, without the stacktrace where smarty was actually involved). 

Doing `throw new Exception("...reason..", null, $e)` will keep the original stacktrace, and allow adding messages like "fetch failed", "compile failed", etc. This should allow easier debugging, as the developer will know where this exception was cached/thrown. 

The downside seems to be that all exception will get re-mapped to the base Exception class, so this will not allow doing

```
try {
   $smarty->fetch("abc.tpl");
} catch( Exception1 $e) {
   // handle exception
} catch( Exception2 $e) {
   // handle exception
}

```

wisskid:
@quamis I think you are right, people will be depending on specific Exception classes, so we shouldn't change that.

PR #507, opened 2018-11-07
chmod with corrent umask

ihortymoshenko:


libs/sysplugins/smarty_internal_runtime_writefile.php
```diff
@@ -29,20 +29,14 @@ public function writeFile($_filepath, $_contents, Smarty $smarty)
     {
         $_error_reporting = error_reporting();
         error_reporting($_error_reporting & ~E_NOTICE & ~E_WARNING);
-        $_file_perms = property_exists($smarty, '_file_perms') ? $smarty->_file_perms : 0644;
```
IgorTimoshenko: _file_perms property does not exist, the __get method always returns null for this property as _file_perms declared as obsoleted property

libs/sysplugins/smarty_internal_runtime_writefile.php
```diff
@@ -29,20 +29,14 @@ public function writeFile($_filepath, $_contents, Smarty $smarty)
...
-        $_dir_perms =
```
IgorTimoshenko: _dr_perms property does not exist, the __get method always returns null for this property as _dir_perms declared as obsoleted property

libs/sysplugins/smarty_internal_runtime_writefile.php
```diff
@@ -89,11 +83,8 @@ public function writeFile($_filepath, $_contents, Smarty $smarty)
             error_reporting($_error_reporting);
             throw new SmartyException("unable to write file {$_filepath}");
         }
-        if ($_file_perms !== null) {
```
IgorTimoshenko: At this point $_file_perms always is not null

libs/sysplugins/smarty_internal_runtime_writefile.php
```diff
@@ -89,11 +83,8 @@ public function writeFile($_filepath, $_contents, Smarty $smarty)
...
-            // set file permissions
-            chmod($_filepath, $_file_perms);
-            umask($old_umask);
-        }
+        // set file permissions
+        chmod($_filepath, 0666 & ~umask());
```
IgorTimoshenko: Fixed chmod mode to apply the umask correctly

AnrDaemon: If I were you, I'd simply remove all calls to chmod()/chown(). This is not a file management code, and unless you do something drastic, you should be able to read files you've created. Meaning, no chmod required.

If there's still a problem, then fix your FS settings. All modern FS supports ACL, you can get results much better than ancient POSIX permissions.

AnrDaemon: Again, both your cases fall into "fix your FS permissions already" category.

```
setfacl -Rm "d:u:builder-user:rwX,d:u:fpm-user:rwX,d:m::rwx" -- /templates/directory
```

IgorTimoshenko: @AnrDaemon,

setfacl is one of the ways to solve the issue with permissions.

I suggest to improve work with chmod and add the ability to influence on chmod which is hardcoded

Nimloth: @AnrDaemon
Please also note that it is possible to set the default umask for php-fpm, for instance.
I usually do that and set umask to 0027.

smarty will nonetheless always create cache files with 0644.
Having explicitly set a umask, my expectation is that this default is honored and used.

It is certainly true that you can fine-tune things using ACLs in ways file permissions don't allow for. However, dismissing POSIX as ancient and thus something that can be ignored is in and of itself a rather disturbing point of view, to be frank.

wisskid:
Okay, I've created #772 to at least remove the code that checks Smarty's obsolete properties. Now remains that question of why Smarty (seemingly intentionally) disables the current `umask` setting when writing files and restores the setting afterwards. I can trace it back to a commit 7 years ago by @uwetews, but there is not specific explanation for doing this. I'm a little reluctant to change this without grasping the consequences.

ihortymoshenko:
@wisskid,

I see no reason to do this.

I think it's better to create a directory with permissions `777` and files with permissions `666 - umask`.

See how it's done in Twig [https://github.com/twigphp/Twig/blob/3.x/src/Cache/FilesystemCache.php#L50](here) and [https://github.com/twigphp/Twig/blob/3.x/src/Cache/FilesystemCache.php#L62](here).

AnrDaemon:
I see no reason to touch permissions at all.
Better don't touch permissions at all without a very specific need for that. Since we are not a file manager, it is not our area of responsibility.
I see why we would nee to specify 0777 for `mkdir()` call, but trying to explicitly `chmod()` at best make no difference, and at worst - break the permissions set by the administrator.

ihortymoshenko:
@wisskid,

I've described the issue [here](https://github.com/smarty-php/smarty/pull/507#discussion_r383986208), it's a real case, you cannot call the `chmod` function to fix permissions with `umask` without specifying some permissions (666).

AnrDaemon:
I already said that your deployment script should have been executed compilation from web user or have fixed permissions after compilation. Either way, not really a Smarty issue.

ihortymoshenko:
@AnrDaemon,

Okay, as you wish, 4 years is too long for this question.

PR #522, opened 2018-12-11
Support of count in 7.2 (old behavior)

welrachid:
Support of count in 7.2 (old behavior).
just inserted a conditional if that checks if count parameter is array or instanceof countable. if not it checks if it is null then it returns 0 otherwise it always returns 1

### DIFF
--- a/libs/Smarty_Compiler.class.php
+++ b/libs/Smarty_Compiler.class.php
@@ -1200,7 +1200,7 @@ class Smarty_Compiler extends Smarty {
             $output .= "    foreach (\$_from as $key_part\$this->_tpl_vars['$item']):\n";
             $output .= "        {$foreach_props}['iteration']++;\n";
         } else {
-            $output .= "if (count(\$_from)):\n";
+            $output .= "if (count((array)\$_from)):\n";
             $output .= "    foreach (\$_from as $key_part\$this->_tpl_vars['$item']):\n";
         }
         $output .= '?>';


welrachid:
Hi. I just did a new commit regarding casting of StdObj in a foreach so that count($_from) where $_from is a stdobj is now being casted before count, so that it doesnt make a warning about countable.

wisskid:
Hi @welrachid
I'm sorry you haven't heard from use earlier. I've recently joined the Smarty project and am trying to get it moving again. Do you still really need this? I'd rather stop supporting Smarty2.
Whoever still uses it is probably better served with stability...
Ok?

welrachid:
Hi @wisskid Thanks for reaching out. Imade those changes for myself because i have quite a few projects that needed to get the PHP version up to a newer. Since i didnt want to rewrite everything in the project (because all the code already was more or less php7.2 compliant) i needed a way for my array counts to stop making warnings and (potential fatals). If it's no suit for the smart2 project since it is discontinued its ok with me. i just wanted to contribute with the time i already spent :)

again. thanks for reaching out.

libs/Smarty_Compiler.class.php
```diff
@@ -1955,7 +1955,11 @@ function _parse_modifiers(&$output, $modifier_string)
                 $_modifier_args = '';
 
             if ($_map_array) {
-                $output = "((is_array(\$_tmp=$output)) ? \$this->_run_mod_handler('$_modifier_name', true, \$_tmp$_modifier_args) : " . $this->_compile_plugin_call('modifier', $_modifier_name) . "(\$_tmp$_modifier_args))";
+                if(strtolower($_modifier_name) == 'count'){
+                    $output = "((is_array(\$_tmp=$output)) ? \$this->_run_mod_handler('$_modifier_name', true, \$_tmp$_modifier_args) : ((is_array(\$_tmp) || \$_tmp instanceof Countable)?" .  $this->_compile_plugin_call('modifier', $_modifier_name)."(\$_tmp$_modifier_args):(is_null(\$_tmp)?0:1)))";
```
wisskid, line 1959: Won't `is_array(\$_tmp) || ` always be false here?

wisskid:
If you put it that way :)
Your changes look great and I can't think how they'd hurt anyone. I've added a single comment. Thanks for you help!

libs/Smarty_Compiler.class.php
welrachid: Not sure if i understand your question. So the idea here is just that i make a check for the $_tmp to be either array or instance of countable. if this is the case we call normally, but if both of them are false we a second check is_null(\$_tmp)?0:1
so that we return the number 0 if $_tmp is null and otherwise its 1.

welrachid: and sorry for the late response. you've been on my list and corona unfortunatly made us go into "war mode" at the company ;)

welrachid: Sorry didnt see i had to click resolve conversation (havent made any pull requests before)

PR #527, opened 2018-12-18
Test against PHP 7.3

JayBizzle:


wisskid:
PHP7.3 is already in .travis.yml

Issue #336, opened 2017-02-22
modifier default not working with @ suppression 


alabama:
The default modifier just access the tpl_vars array with the non existing key and tries to suppress the error outcome with '@' like:
`$tmp = @$_smarty_tpl->tpl_vars['bShowInfo']->value`

This won't work in PHP >= 5.4 (5.3 as well?)
There should be at least an "array_key_exist | isset" check to get the functionality right

Miksser:
Is it possible to get rid of suppression in 'default'? 
This behavior is very disturbing during debugging through xDebug =(

Miksser:
Is it possible to get rid of suppression in 'default'? 
This behavior is very disturbing during debugging through xDebug =(

wisskid:
@Miksser @antman3351 I've added an empty check for variables for `|default` modifier in #650 . It does not appear to break anything. Can you confirm this fixes your problem?

antman3351:
I just had a quick look at the change, so I might have missed something, but doesn't this change the behaviour of the default modifier?
before default was used for unset variable, null or an empty string. Using empty() default would also be applied to false, 0 , '0' and [ ].

wisskid:
Well spotted, @antman3351 ! Not sure how that got past me. I thought I ran the new unit tests against the old code (seeing no change in behavior) but I must have somehow messed that up. 
Added the 0, '0' and [] cases to the unit tests and fixed the implementation.
Is this better?

wisskid:
The default modifier has already been changed to longer use the @ modifier in https://github.com/smarty-php/smarty/pull/629

Issue #499, opened 2018-10-17
smarty isset not working notice Undefined offset 3.1.32v and 3.1.33v


jvalencia74:
when trying to update it happens to me in both versions and it is in a section loop 

{section name=field loop=$result} 
{if ((isset($result[field].changeState)) and ($role_update == "1"))} 
{else} 
{/if} 
{/section} 

template return error 

smarty\ce7fa0ecf0a71469a5f7beae553a6f8123a19440_0.file.Maintenance.Dispatchers.input.tpl.php:203] Undefined index: changeState 

it only happens when reading an array with section 

I have the same error in several templates with different index 

works well in 3.1.30v

tks

mylemans:
I have the same issue, almost (if not all) my sections broke!

pszalko:
I can confirm this issue for me too. Smarty code inside `{section}` does not generate `isset()`.

jeckyl:
Hi,
I also confirm this concern. Quite embarrassing in the development of testing its variables to realize that the test itself does not work.

tmcreativeltd:
Come on Smarty... surely this is what isset is for? Any plans on a fix?

pszalko:
I have encountered this issue again...

Here is some follow up if and workaround that I found.

**Use case:**
The following Smarty code raised PHP Notice:
```
<li class="{if isset($availabilityDates[$group->getId()][$dateString]) == true} selected{/if}"
```

After short inspection:

**1. Investigation:**
I have found out that this issue might be related to implementation in `smarty_internal_templatecompilerbase.php:633` (for Smarty v3.1.33):
```php
$pa[] = (strpos($p, '(') === false) ? ('isset(' . $p . ')') : ('(' . $p . ' !== null )');
```
Variable `$p` holds the exact line of the template that is being parsed by Smarty.
In that particular line my template code is `isset(... [$group->getId()]...)`.
Do you see already what's going on? Smarty looks for `(` char, and if found `isset(...)` is stripped and replaced by `!== null` in compiled template.

**2. Workaround:**
Remove any parenthesis from argument of `isset()` method.

Here is workaround implementation for my use case:
```
{assign var="groupId" value=$group->getId()}
<li class="{if isset($availabilityDates[$groupId][$dateString]) == true} selected{/if}"
```

I hope this knowledge will help someone in Smarty team to resolve this issue.

wisskid:
Duplicate of #453

Issue #498, opened 2018-10-09
isset on array


beaceberq:
`isset` is not generated when we specify the object's property / method as an index

Example:
Code on .tpl:
`{if isset($specialOffers[$category->getId()])}gray{/if}`
Generated code:
`<?php if ((($_smarty_tpl->tpl_vars['specialOffers']->value[$_smarty_tpl->tpl_vars['category']->value->getId()] !== null ))) {?>gray<?php }?>`

pszalko:
Any update on this issue?

nenetchi:
+1 for backticks substitution in array index

- PHP 7.0.33/7.3.1
- Smarty 3.1.33

Template
```
{if isset($foo["prefix_`$bar`"])}
1
{/if}
```

will generate

```php
if ((($_smarty_tpl->tpl_vars['foo']->value["prefix_".((string)$_smarty_tpl->tpl_vars['bar']->value)] !== null ))) {?>
1
<?php }
```

with `isset` missing

pszalko:
I believe this is similar issue to #499 
and I hope the workaround from https://github.com/smarty-php/smarty/issues/499#issuecomment-471963311 will help.

Anyway, I hope to see it resolved soon.

wisskid:
Duplicate of #453

PR #567, opened 2020-01-27
This appears to fix #453, #498, #499 and possibly more issues. 

### LINKED ISSUE #453: isset() check fails within section $VAR[loop].key syntax if "key" or "loop" does not exist
Hello there!

I recently updated the Smarty library in a project from v3.1.30 to v3.1.32 and noticed, that this version seems to have a problem with the `isset()` check if a var is not set and I'm using the syntax you usually use to address something in a section with `$VAR[counter].subvalue`. As soon I roll back to v3.1.30 the same template code works again so I think it must have to do something with a change made in 3.1.31 or 3.1.32. I'm using both with PHP7.2 and error_reporting is set to strict including all notices for uninitialized vars in my dev environment.

Has the syntax just changed or does `[]` fail now due to a bug? (It worked on 3.1.30).

**I made an example for you here:**

```
- PAGE START
{section name=topics loop=$TOPICS}
  {if isset($TOPICS[topics].thisoneexists)}
     this one exists
  {/if}

  {if isset($TOPICS[topics].thisdoesnotexist)}
     this one does not exist
  {/if}

  {if isset($TOPICS[bla].blubb)}
     this doesn't even nearly exist ;)
  {/if}
{/section}
- PAGE END
```

With Smarty v3.1.30 I see:
```
- PAGE START
this one exists
- PAGE END

```

so this is the expected behaviour: The vars that are not existing in the section array fail the isset check, but the template gets compiled further to the end.

However on Smarty v3.1.32 I see:
```
- PAGE START
this one exists
```

In other words: As soon I have an isset with a non-existing array key the template stops to compile further and I end up just seeing a blank page. No message in the error log, no message when I turn on Smarty Debugging ... very mysterious.

~~The funny thing is when I write it like this `{if isset($TOPICS.topics.thisdoesnotexist)}` it seems to work.~~
> **EDIT:** Correction, this is also no option to solve the problem! I just noticed the method with dots doesn't work in sections. This will always return a false isset() check even if a value is set. So it seems I definitely need the `$TOPICS[topics].thisdoesnotexist` Syntax here.

Am I doing something wrong here or is that a bug? It makes me wonder that the syntax with `[]` does NOT fail if the value is really set and only fails when its not set, so if you got rid of the `[]` variant and there is a different syntax for sections, shouldn't this fail with a syntax error in both cases then? Or shouldn't it behave like in 3.1.30 where using that worked correctly and nonexistant values didn't cause the template compiling to stop/fail?

Thanks in advance!

wisskid:
Will need further testing and unit tests.

Using the @-operator will supress errors generated in code wrapped in the `isset()` call. That's not good. Better than what we have now, but not ideal.


wisskid:
Another approach would be to improve the regex that tries to determine whether `$p` is a variable. This might do it:
`^\$[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*((->)[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*|\[[^\]]*\])*$`

wisskid:
Slight improvement to balance out brackets with nested brackets: `^\$[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*((->)[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*|\[.*]*\])*$`

uwetews:
Please note that the issues at handling isset() and empty() calls can get much more complex.
If the variable tested is an object its properties can be variable.
Array indexes may contain variables.
You will get runtime errors when some of this subvariables are undefined.
I was working on a solution before I got ill last year.
I should be able to work on this problem in about 2-3 weeks

wisskid:
@uwetews great to hear you are coming back! I was thinking a truly proper solution would require lexing/parsing the parameter to ensure it's a valid php variable instead of using a regex. But my solution posted above seems pretty effective too.
I'd love to hear your thoughts. Get well soon!

wisskid:
@uwetews a similar fix in `libs/plugins/modifiercompiler.default.php` would fix https://github.com/smarty-php/smarty/issues/336

glensc:
please note that you can't really get by pattern match accurate result whether 
 `empty()` argument is non-variable.

see discussion and example here:
- https://github.com/PHPCompatibility/PHPCompatibility/issues/410

PR #569, opened 2020-01-28
fix: "too many shorthand attributes" error when one attr given

bhaal-nig:


AnrDaemon:
Now, that tests are merged into base package and can be executed straight away, adding tests for your new/fixed features should be a priority, IMO.

PR #572, opened 2020-02-19
Recognize https as a http protocol in {fetch}

### LINKED ISSUE #465: {fetch} with https protocol
`{fetch}` allows us fetching external files via http/https.
In case of http protocol, [fsockopen() is used to fetch the content specified as "$params['file']"](https://github.com/smarty-php/smarty/blob/cc4d8fa1a07e18f00951a921041e37c87015dcb2/libs/plugins/function.fetch.php#L155-L157), but **https** is not.

The `if` statement below allows only http: 
https://github.com/smarty-php/smarty/blob/cc4d8fa1a07e18f00951a921041e37c87015dcb2/libs/plugins/function.fetch.php#L58

The file specified with **https** protocol can [be fetched with file_get_content()](https://github.com/smarty-php/smarty/blob/cc4d8fa1a07e18f00951a921041e37c87015dcb2/libs/plugins/function.fetch.php#L210), but ideally I think it should achieved with fsockopen() just like http.

Is there any reason that **https** is not allowed [here](https://github.com/smarty-php/smarty/blob/cc4d8fa1a07e18f00951a921041e37c87015dcb2/libs/plugins/function.fetch.php#L58)?

wisskid:
Fixes #465

### DIFF
--- a/change_log.txt
+++ b/change_log.txt
@@ -1,4 +1,5 @@
  - fix foreachelse on arrayiterators https://github.com/smarty-php/smarty/issues/506
+ - use same method for handling https as http in `{fetch}` https://github.com/smarty-php/smarty/issues/465
 
 ===== 3.1.34 release ===== 05.11.2019
 13.01.2020
--- a/libs/plugins/function.fetch.php
+++ b/libs/plugins/function.fetch.php
@@ -49,7 +49,7 @@ function smarty_function_fetch($params, $template)
         }
     }
     $content = '';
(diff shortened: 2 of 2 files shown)


wisskid:
@uwetews my travis checks in this PR are failing because https://github.com/smarty-php/smarty/pull/567 hasn't been merged yet. I'm used to having the unit tests in the same repo as the source for this reason. We can exclude them from the git export if size is what worries you?

Issue #578, opened 2020-04-08
{$smarty.template} with "_" characters and custom resource


EDCScott:
This is a followup to issue #490

The original report was that the {$smarty.template} variable reports incorrectly for template names that contain an underscore.  After some further testing, I was able to isolate that the issue only happens when using a custom resource.

I've attached a sample which recreates the issue.  To reduce the size of my upload, I didn't include the smarty installation, which is expected to be in a "smarty" sub directory.

[example.zip](https://github.com/smarty-php/smarty/files/4448043/example.zip)

EDCScott:
I've been able to isolate the issue to the following commit:

https://github.com/smarty-php/smarty/commit/614ad1f8b9b00086efc123e49b7bb8efbfa81b61

EDCScott:
I have the issue resolved locally, but it appears as though I don't have permissions to push to this repository.  Is there a different procedure, other than requesting permissions to push branches and open pull requests?

wisskid:
Hi @EDCScott! At github it is customary for you to fork this project in your own namespace, create a branch that fixes the issue there and than create a pull request (PR) from your repo to this one.

EDCScott:
Done.  Thank you for the insight.

wisskid:
NIce! I'll review the PR asap.

wisskid:
Fix was already merged last year.

Issue #490, opened 2018-09-19
{$smarty.template} with "_" characters


EDCScott:
The {$smarty.template} variable reports incorrectly for template names that contain an underscore.

wisskid:
I don't think it does. I just tested it, and it works fine. If you have any more details, please open a new bug report.

PR #581, opened 2020-04-13
Bug fix for underscore in template name

EDCScott:
This addresses bug #578 and #490, by correcting the {$smarty.template} variable when using template names containing the "_" character along with a custom resource.

This also extends the maximum template name length from 25 to 255 when using a custom resource.


wisskid:
Hi @EDCScott . I've reviewed your changes and I see no problem in allowing for an underscore the regex. The changing of the max length might be a problem, though. The template name is combined with prefixes and postfixes into a cache file name, and the total length of the cache file name probably should exceed 255 chars. So allowing 255 chars for the template name alone might cause problems there.

EDCScott:
@wisskid I agree.  The 255 length was chosen as it's the maximum filename length on many platforms (https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits), including NTFS and various versions of ext.  If we're going to be adding something to that file name, then we should abbreviate it.

Is there a maximum length on the prefix or postfix, or are they user-definable?  If they have a limit, we should probably make the maximum length here be (255 - (max prefix) - (max postfix)).  If there is no maximum length, then that's probably another bug in it's own right.  Perhaps we limit prefix and postfix to 64 characters, and then limit the template name itself to 127 characters?  

Or if you want to make things really robust (and complicated), the limits are technically 255 BYTES.  I'm not sure if Smarty has made an effort to account for multi-byte filenames.

The lower effort changes would probably be just to adjust the 255 to 127 and call it a day.  If smarty isn't providing proper error handlers for overlength files, then it's not really a different developer experience if you overflow the maximum filename length, than if it's unsuspectedly trimming your filename length internally.

Your thoughts?

AnrDaemon:
Since

1. cache ID is user-defined (read: arbitrary length), 
2. cache location is user-defined (read: can be anything, not just filesystem),

the entire discussion about lengths of the names is questionable.

But I see a different problem with this PR: it lacking tests.

BarbzYHOOL:
looks fine, maybe repo maintainers can just edit the character limit to their liking and merge?

libs/sysplugins/smarty_resource_custom.php
```diff
@@ -88,6 +88,6 @@ public function getContent(Smarty_Template_Source $source)
      */
     public function getBasename(Smarty_Template_Source $source)
     {
-        return basename(substr(preg_replace('/[^A-Za-z0-9.]/', '', $source->name), 0, 25));
+        return basename(substr(preg_replace('/[^A-Za-z0-9._]/', '', $source->name), 0, 255));
```
AnrDaemon, line 91: I'd wrap this transformation into a separate method.
You already using it in two places and the code like this is untestable.

Issue #364, opened 2017-05-12
Support: looking a way to get smarty parse tree


glensc:
I'm looking a way to get tokens from smarty, something similar to 
[getTags()](http://www.smarty.net/docs/en/api.get.tags.tpl), but with other tokens as well like text nodes.

The intention is  to write external parser to extract `{t attrib="123" attr2='foo bar'}text here{/t}` as tokens by smarty own parser, instead of using regexps on source template file.

Here's example how `getTags` is being used good way externally:
https://github.com/geekwright/Po/blob/v1.0.4/src/PoInitSmarty.php#L189-L218

however that solution does not apply to me because i need the TEXT element from between `t` and `tclose` notes.

AnrDaemon:
I don't quite understand, why YOU want to parse Smarty template?
Why don't you just write a Smarty plugin and Smarty will send everytihing you you ready and parsed?

glensc:
see related links in this issue. i need `{t opts=111}...{/t}` not render actual template.

can you provide example how to get same result with plugin? what is plugin anyway? any docs?

AnrDaemon:
Please read http://xyproblem.info/ before answering my question.
Now, the question: WHY do you want to do this?
What would be the end result for you?

glensc:
looks like you never visited any links given. sad.

[smarty-gettext](https://github.com/smarty-gettext/smarty-gettext) has tool named [tsmarty2c](https://github.com/smarty-gettext/smarty-gettext/blob/1.5.1/tsmarty2c.php) (manual: [tsmarty2c.1](https://github.com/smarty-gettext/smarty-gettext/blob/1.5.1/tsmarty2c.1)), which uses regexps to scan templates for the translations. there are several bugs and features requested ([#1](https://github.com/smarty-gettext/smarty-gettext/issues/1), [#15](https://github.com/smarty-gettext/smarty-gettext/issues/15), [#20](https://github.com/smarty-gettext/smarty-gettext/issues/20)) which i [tried to solve](https://github.com/smarty-gettext/smarty-gettext/pull/16#issuecomment-300573040) with [some code refactoring](https://github.com/glensc/smarty-gettext/commit/9f759cb96ac4bf774655f5d7da5e5d46338f96d0) but decided that that i'll be doing it wrong inventing all moving parts myself and rather use existing components.

[...]

PR #582, opened 2020-04-13
Bugfix/strip doesnt remove whitespaces after comment

wisskid:


glensc:
This is regression to the tsmarty2c tool:
- https://github.com/smarty-gettext/tsmarty2c/pull/16

Altho I'd prefer if smarty provided API to walk over nodes:
- https://github.com/smarty-php/smarty/issues/364

PR #585, opened 2020-04-14
Allow/Deny inline constants

### LINKED ISSUE #554: Direct access to php constants breaking smarty
Hi,

We upgraded from Smarty 3.1.21 to 3.1.33

Apparently, if you have a php variable defined as follows :

`define("myvar", "My var value");`

It can be accessed in the template as `{myvar}`

Can we stop this behavior ? 

In our app we have a php constant defined named "assign" which breaks the smarty template when the assign tag is used.

There should be a way to not directly "see" these keywords as constants since anyone can break smarty by defining a php variable having the same smarty tag name. And btw the generated exception doesn't say much, it was hard to find.

Thanks !

artcs:
Adding var `$allow_inline_constants` to SmartyClass and using it in templateparser. This way you can globally deny use of inline constants without `$smarty.const.` prefix to avoid various ambiguities. Just set `$Smarty->allow_inline_constants = false`. Defaults to true for backwards compatibility.
Closes #554

### DIFF
--- a/lexer/smarty_internal_templateparser.y
+++ b/lexer/smarty_internal_templateparser.y
@@ -154,6 +154,7 @@ class Smarty_Internal_Templateparser
         $this->template = $this->compiler->template;
         $this->smarty = $this->template->smarty;
         $this->security = isset($this->smarty->security_policy) ? $this->smarty->security_policy : false;
+        $this->allow_inline_constants = $this->smarty->allow_inline_constants;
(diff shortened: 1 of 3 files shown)


AnrDaemon:
Missing tests.

artcs:
I found function calls like `{if strpos(…) !== false}` broken with my first version, so I changed it a bit. But this is only speculative trial and error, cause I honestly don't know exactly how the templateparser works. So this should definitely be reviewed by someone, who is familiar with the parser syntax. This one could then also write the corresponding tests.

PR #589, opened 2020-05-04
Removed bogus calls to addslashes

AnrDaemon:


wisskid:
what issue are you fixing here? I can see what you are doing, but is there an actual failing use case or security consideration?

AnrDaemon:
The same issue as if you would be using `htmlspecialchars()` to escape SQL.
`addslashes()` has nothing to do with escaping PHP strings for code generation. `var_export()` is specifically designed for that.

wisskid:
As @m-ober said, this seems quite OK. All usages are string interpolation. I'm sure AnrDaemon's suggestion works fine too, but why change a winning team?

PR #590, opened 2020-05-04
Fixed replace modifier by converting encoding if needed

AnrDaemon:
`mb_split` will fail if `$pattern` or `$string` contains byte sequences not valid for `mb_regex_encoding()`.

Convert both strings and set regex encoding before calling `mb_split()` if needed.

Fixes smarty-php#549

Issue #353, opened 2017-04-24
Question: Smarty 3.2


ghost:
Hello,

Almost 4 years ago you posted something about smarty 3.2 being created, do you have any idea when this new version of smarty will be published? Or do you perhaps already have a dev version for this which simply hasn't been posted yet?

Sincerely yours,

flexjoly:
Hi, 
I have the same question
Maybe we can start a fork to develop a follow-up, since smarty -development seems to be death and uwetews is not responding (much).

Smarty needs to be updated for php 7 and must start using namespaces and the errorhandling must be updated. 

Any volunteers?

Greetz, flexjoly

pinkminds:
We run smarty on PHP 7, so no problem with that. But it would indeed be good to know if the development of smarty is still active and by when next releases are scheduled...

uwetews:
Hallo
Smarty does run without Problems on php 7.
I myself are recovering from a longer illness and should be active very
soon again.

Regards Uwe


pinkminds <notifications@github.com> schrieb am Mo. 19. Juni 2017 um 13:12:

Gemorroj:
@uwetews, wish you good health 💊

pinkminds:
@uwetews, get better soon!

PoppedWeb:
@uwetews, I hope you will be better soon! I know I am asking a lot, but could you give us like some sort of update on the progress of smarty 3.2? Because we all are very curious as to what you are planning to do with smarty 3.2!!!

uwetews:
Smarty 2 version 2.6.28 is the latest version which does support PHP 4.
However if you want to use composer it does require as minimum PHP 5.3.

PHP 4 support has been dropped in Smarty 2 version 2.6.29 and higher.

Smarty 2 version 2.6.29 and higher is compatible with PHP versions:
5.2 and higher
7.0
7.1
and even the upcomming 7.2

Corresponding PHPUnit tests will be published i the next couple of days.

For Smarty 3.1 I have tested PHP 7.2 compatibility back to version 3.1.24, but I think earlier version should also run.

PHPUnit tests for PHP 7.2 have already been added to the 3.1.32-dev version of Smarty 3.1.

I have dropped the development of Smarty 3.2, but working now on Smarty 4.

I will release a feature list in the near future.

flexjoly:
I am very happy to read this!! Thanks @uwetews ! 
Can we help we development?

jcfalcone:
[...]

PR #591, opened 2020-05-04
PHP 5.3 compatibility fixes

AnrDaemon:


AnrDaemon:
@wisskid tck-tck-tck…

wisskid:
Ouch

AnrDaemon:
In this light, I'd like to raise https://github.com/smarty-php/smarty/issues/353#issuecomment-615367537 once more.

Stuff moving forward and supporting highly outdated engine versions becoming increasingly problematic.

I've had to spend half an hour just to be able to successfully run tests using 5.3. AFTER the moment I fixed linting errors. I'm not sure it can be run on 5.2 anymore. (Though, Composer only supports 5.3.2+, so the point is probably moot.)

wisskid:
I was afraid you’d bring that up now. Now, I don’t see why we’d have to support 5.3 or older anymore. It’s been EOL almost six years. But dropping php5.6 could cause some actual pain, right?

And, jus being curious, how did you notice this? Are you still running 5.3?

AnrDaemon:
EOL or not, Smarty declares compatibility (right now, with 5.2+, but in fact I doubt it can run on anything older than 5.3) with certain PHP version. Changing minimum engine version and retain the same library version is a bad service, IMO.

This is why I suggest changing version number to move things forward. We can keep 3.1/5.3 branch in case something crop up that would require urgent attention, but other than that, I see only benefits.

Even language defaults were changed long before 7.x. Just look at the recent #590. Not that the code could have been shorter, the final version took much care about preserving set encodings where possible, but default settings for both mb_regex_encoding and mb_internal_encoding were changed not once since 5.3.

No, I'm not using 5.3 daily, but I keep it fully operational for regression tests.

wisskid:
I totally agree. When we start dropping compatibility with certain php versions, we should up the version number. And we should get unit tests running in travis for each php version that we support.
But if we can easily support php5.6, then I don't see any harm in doing so, right?

AnrDaemon:
Let's move it to a separate issue, shall we?

PR #599, opened 2020-06-19
Update to HTML5-syntax in debug template

JonisoftGermany:
* Replace deprecated tags
* Use CSS3 selectors
* Update html-frame
* Remove deprecated syntax
* Harmonize syntax in general


scottchiefbaker:
This seems like a reasonable upgrade to me. Unless there is some specific reason to keep the template XHML, I think this is merge worthy.

lkppo (approved):

PR #604, opened 2020-07-29
Set $errcontext argument optional to support PHP 8

xorti:
- Argument is optional and deprecated in PHP 7.2

### DIFF
--- a/libs/sysplugins/smarty_internal_errorhandler.php
+++ b/libs/sysplugins/smarty_internal_errorhandler.php
@@ -65,7 +65,7 @@ class Smarty_Internal_ErrorHandler
      *
      * @return bool
      */
-    public static function mutingErrorHandler($errno, $errstr, $errfile, $errline, $errcontext)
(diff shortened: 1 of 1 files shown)


wisskid:
Why not remove it all together?

libs/sysplugins/smarty_internal_errorhandler.php
```diff
@@ -65,7 +65,7 @@ public static function muteExpectedErrors()
      *
      * @return bool
      */
-    public static function mutingErrorHandler($errno, $errstr, $errfile, $errline, $errcontext)
+    public static function mutingErrorHandler($errno, $errstr, $errfile, $errline, $errcontext = [])
```
glensc, line 68: this is not valid for php < 5.4

xorti:
@wisskid I don't want to break the compatibility with earlier versions of PHP, smarty's latest version still supports >=5.2.

Removing it, might not break the house, but making it optional sounds safer.

wisskid:
Thank you, I've included this in https://github.com/smarty-php/smarty/pull/608 fixing several other php8-issues in one go.

PR #608, opened 2020-09-10
Fixes for php8.0.0beta3

wisskid:


wisskid:
@AnrDaemon care to review?

.travis.yml
```diff
@@ -1,38 +1,44 @@
...
+install:
+  - travis_retry composer require phpunit/phpunit:^7.5
 
-matrix:
+jobs:
   include:
-   - php: 5.3 # Composer requires PHP 5.3.2+ to run, so we cannot test below 5.3
+   - php: 5.3 # Composer and PHPUnit require PHP 5.3.2+ to run, so we cannot test below 5.3
      dist: precise # PHP 5.3 is supported only on Precise.
+     install: travis_retry composer require phpunit/phpunit:^4.8
```
AnrDaemon, line 12: This is unnecessary, as composer would automatically select best matching version.
See https://github.com/smarty-php/smarty/blob/master/composer.json#L43

AnrDaemon: If you actually want to change something to ensure version choices, change `config.platform`. The rest should be done automatically.

wisskid: Thank you for your input! I agree your suggestion is better. I remember running into issues with travis builds failing because it was picking the wrong phpunit version. But I'm not sure why, so I'll try again.

wisskid: Well, that worked beautifully. Not sure what went wrong earlier. Thanks again!

AnrDaemon: Ordering matters. If first specified version matches current constraints, Composer won't try the next one.

libs/sysplugins/smarty_internal_compile_private_special_variable.php
```diff
@@ -94,9 +94,9 @@ public function compile($args, Smarty_Internal_TemplateCompilerBase $compiler, $
                         break;
                     }
                     if (strpos($_index[ 1 ], '$') === false && strpos($_index[ 1 ], '\'') === false) {
-                        return "@constant('{$_index[1]}')";
+                        return "defined('{$_index[1]}') ? constant('{$_index[1]}') : null";
```
shopx, line 97: `return "defined('{$_index[1]}') ? constant('{$_index[1]}') : null";`
change to
`return "(defined('{$_index[1]}') ? constant('{$_index[1]}') : null)";`
??

AnrDaemon: Useless braces are useless.

AnrDaemon: I missed the fact that this is a code generation, in which case the braces may be necessary, given the fact the resulting expression may be used in ambiguous context..

Issue #510, opened 2018-11-15
filemtime(): stat failed for warnings


blaaat:
If I run a `rm -Rf templates_c/*` these warnings appear. 

`is_file` calls are cached; so the checks surrounding filemtime( are not conclusive. 
I don't want to use the deprecated `muteExpectedErrors` to block all smarty warnings. 

Shouldn't the @ error supression on filemtime() come back? or lower the error_reporting level during these calls?

> 
> .../vendor/smarty/smarty/libs/sysplugins/smarty_template_resource_base.php:148Smarty_Template_Resource_Base::getTimeStamp	
> .../vendor/smarty/smarty/libs/sysplugins/smarty_template_compiled.php:177Smarty_Template_Compiled::compileTemplateSource	
> .../vendor/smarty/smarty/libs/sysplugins/smarty_template_compiled.php:150Smarty_Template_Compiled::process	
> .../vendor/smarty/smarty/libs/sysplugins/smarty_template_compiled.php:105Smarty_Template_Compiled::render	
> .../vendor/smarty/smarty/libs/sysplugins/smarty_internal_template.php:216Smarty_Internal_Template::render	
> .../vendor/smarty/smarty/libs/sysplugins/smarty_internal_template.php:385Smarty_Internal_Template::_subTemplateRender	
> ..../templates_c/b32824a85ca6c72e557af4860bf7fc86118fbde2_0.file.xxx.html.php:37content_5bed2640879fd7_75737843	
> .../vendor/smarty/smarty/libs/sysplugins/smarty_template_resource_base.php:123Smarty_Template_Resource_Base::getRenderedTemplateCode	
> .../vendor/smarty/smarty/libs/sysplugins/smarty_template_compiled.php:114Smarty_Template_Compiled::render	
> .../vendor/smarty/smarty/libs/sysplugins/smarty_internal_template.php:216Smarty_Internal_Template::render	
> .../vendor/smarty/smarty/libs/sysplugins/smarty_internal_templatebase.php:232Smarty_Internal_TemplateBase::_execute	
> .../vendor/smarty/smarty/libs/sysplugins/smarty_internal_templatebase.php:134Smarty_Internal_TemplateBase::display

wisskid:
I would expect to see warnings from running processes when I delete cache files they rely on during execution in another process, so I don't really think this qualifies as a bug?

blaaat:
This issue was due to the opcache.enable_file_override ini setting. 

I did not realize directadmin changes the default for this setting.

PR #614, opened 2020-11-19
Bugfix for #510 filemtime(): stat failed for warnings

stephanlueckl:
in case of an active opcache the file exists, but the filemtime method can't find the file.
A solution is to set the timestamp to the actual time()


dmetzner (approved):
The suggested solution looks good to me!

@wisskid Would it be possible to add this small fix to the next release? Thx!

wisskid:
Closing this as #510 is already closed as 'not a bug'.

PR #618, opened 2020-12-05
`const` condition fix

### LINKED ISSUE #609: Big issue on a simple IF condition
Big issue on a simple if condition:

```
{if 1 or $smarty.const.IS_PHONE}
   NOT HERE
{else}
  ARRIVE HERE !!! Strange, no ?
{/if}
```

Seems that not arrive in conition {if 1 or $smarty.const.IS_PHONE}

### LINKED ISSUE #613: smarty.const wrongly evaluated for usage with comparison operators
This refers to the current fixes for PHP 8 see #605 
simple dummy example
somewhere in php `define('_THEME', 'light');`
template

`{if $smarty.const._THEME == 'dark'}
	dark mode selected 
{elseif $smarty.const._THEME == 'light'}
	light mode selected 
{/if}`

expected output
`light mode selected`

but using current master with the PHP 8 fixes output is
`dark mode selected`

this happens due to generated php
see file attached
[example.txt](https://github.com/smarty-php/smarty/files/5389662/example.txt)

ok, looks like the in #605 proposed change for constant handling was already built in
imo the way its done is wrong 
and, and its seems to me that smarty should not be responsible for handling undefined constants
I think this should be done at the application layer in PHP or whatever
if an constant is not defined and the application developer knows that this can happen (should not) the developer is responsible to handle this situation
smarty should not hide this

liborm85:
`const` condition fix introduced in PR #608.

Fixes #609
Fixes #613

### DIFF
--- a/libs/sysplugins/smarty_internal_compile_private_special_variable.php
+++ b/libs/sysplugins/smarty_internal_compile_private_special_variable.php
@@ -94,9 +94,9 @@ class Smarty_Internal_Compile_Private_Special_Variable extends Smarty_Internal_C
                         break;
                     }
                     if (strpos($_index[ 1 ], '$') === false && strpos($_index[ 1 ], '\'') === false) {
-                        return "defined('{$_index[1]}') ? constant('{$_index[1]}') : null";
+                        return "(defined('{$_index[1]}') ? constant('{$_index[1]}') : null)";
                     } else {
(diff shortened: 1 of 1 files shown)


liborm85:
@wisskid @AnrDaemon Can you merge this fix? And release new version?

Issue #623, opened 2021-01-04
Chain @return types


blaaat:
For Chaining `->assign('x',1)->displa('x');` and IDE/static analysis purposes can I suggest having the proper @return docblock?

so for example: 
`     * @return static current Smarty_Internal_Data (or Smarty or Smarty_Internal_Template) instance for
     *                              chaining`
  
instead of: 

`    * @return Smarty_Internal_Data current Smarty_Internal_Data (or Smarty or Smarty_Internal_Template) instance for
 chaining`

wisskid:
@blaaat could you provide a PR?

PR #627, opened 2021-01-06
#623 Static return type in docblocks; for better IDE support chaining

blaaat:
see #623


wisskid:
@blaaat Wouldn't `@return $this` be more appropriate here?

blaaat:
I don't think so; static is a PHP8+ return type; which could be added as regular return type when php7 support is dropped.

wisskid:
But we're not dropping php7 support, right?

blaaat:
> But we're not dropping php7 support, right?

This is just a doc block; so it's not dropping php7 support.

Issue #630, opened 2021-01-13
Nested Foreach attribute behavior changed


dphadtare:
Smarty **foreach** attribute behavior is changed in the latest version that is 3.1.37.

Nested foreach behavioral change is observed in the latest version(Not sure from which version this has been changed but it is working as expected in version 3.1.21). 
Issue is, after the first iteration of parent foreach, item attribute of child foreach value does not update, to prove the behavioral change refer to the below snippet.

```
{$arrCounter = [ "FirstItem" => [ "First 1", "First 2", "First 3", "First 4" ], "SecondItem" => ["Second 1", "Second 2", "Second 3", "Second 4"] ]}

{foreach item=value1 from=$arrCounter}
	{foreach item=value from=$value1}
            {$value}
	{/foreach}
        <br>
{/foreach}

Value: {$value}
```
**The output of above code is:** 

```
First 1 First 2 First 3 First 4
Second 1 Second 2 Second 3 Second 4
Value: First 4
```

**Expected Output is:**
```
First 1 First 2 First 3 First 4
Second 1 Second 2 Second 3 Second 4
Value: Second 4
```
The expected $value is **Second 4** but it giving a value from the first iteration only.

Not sure what is cause for this behavioral change and did not found any updates/note in the changelog as well.

wisskid:
That is interesting, I'll see if I can find where it originates.

ophian:
How about changing L~267
```php
        $output .= $foreachCompiler->compileRestore(1);
```
to
```php
        $level = $restore > 0 ? $restore : 1;
        $output .= $foreachCompiler->compileRestore($level);
```
in sysplugins/smarty_internal_compile_foreach.php
?
Would that have any downsites elsewhere?

wisskid:
Caused by a change in v3.1.28

wisskid:
> How about changing L~342
> 
> ```
>         $output .= $foreachCompiler->compileRestore(1);
> ```
> 
> to
> 
> ```
>         $levels  = (!is_null($restore) && $restore > 1) ? $restore : 1;
>         $output .= $foreachCompiler->compileRestore($levels);
> ```
> 
> in sysplugins/smarty_internal_compile_foreach.php
> ?
> Would that have any downsites elsewhere?

This doesn't work. `$restore` appears to be 1 anyway (i.e. in my testcase)

wisskid:
I have a fix:
```
[...]

PR #636, opened 2021-01-22
No longer save item variable before foreach loop

wisskid:
so it will not be restored afterwards.
See #630


AnrDaemon:
Missing tests. ;)

PR #644, opened 2021-02-24
plugins: shared: mb_str_replace: mb_split special case fix

kornrunner:
`mb_split` might return false, and following call to `count` throws a `TypeError` in PHP 8 if supplied argument is not of type Countable|array


wisskid:
This has already been fixed in https://github.com/smarty-php/smarty/pull/687

PR #646, opened 2021-03-04
A new option for smarty config: $smarty->outputMinify(true) to result minified html code (what do now the {strip} tpl wrapper)

yourchoice:
Instead to use {strip} wrapper in each template to dispaly a minified html code, will be nice to have something like:
"$smarty->outputMinify(true);"

In config to can write:
```
define('DEBUG', true);
$smarty->outputMinify(DEBUG ? true : false);
```
It's very hard now to switch between minified and normal code source (html) !


wisskid:
This should not be a PR. Feature requests can be done through https://github.com/smarty-php/smarty/issues

PR #647, opened 2021-03-12
Optimize away two stat calls per template request

wisskid:
…assuming the source template exists. Note that this implementation disables support for the fallback to templates relative to getcwd() if a template is not found in the specified template dirs, so it might be better to release this as a new version, or to add an alternative resource for this.


wisskid:
This did produce meaningful speed improvement IRL. Closing it.

PR #649, opened 2021-03-21
Naive attempt at stopping incorrect escape type being used

DavidGoodwin:
I have found myself doing "....|escape:html_attr" or" ....|escape:quote" a few times. It'd help if there was some sort of warning logged if an unsupported escape type is used.


DavidGoodwin:
Thanks @wisskid .... I didn't expect it to be so painless :)

PR #629, opened 2021-01-08
Feature/php8 support

wisskid:

PR #650, opened 2021-03-21
default modifier uses empty instead of @ error suppression modifier

### LINKED ISSUE #336: modifier default not working with @ suppression

wisskid:
default modifier uses empty instead of @ error suppression modifier when testing a variable

Fixes #336


wisskid:
The default modifier has already been changed to longer use the @ modifier in https://github.com/smarty-php/smarty/pull/629

PR #651, opened 2021-03-22
Fixed render crash when using inline include with a string template

### LINKED ISSUE #639: Render crash on changing file include modes
```
PHP Notice:  Undefined index: 957e065e65187cb9f8a1b05a74e90f4f1b7188b0 in .../vendor/smarty/smarty/libs/sysplugins/smarty_internal_template.php on line 322
PHP Fatal error:  Uncaught  --> Smarty: plugin Smarty_Resource_ is not a valid name format <--
  thrown in .../vendor/smarty/smarty/libs/sysplugins/smarty_internal_method_loadplugin.php on line 40
```
The reproduction case will follow.

wisskid:
Fixes #639

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - modifier escape now triggers a E_USER_NOTICE when an unsupported escape type is used https://github.com/smarty-php/smarty/pull/649
 
+### Fixed
+- Fixed render crash when using inline include with a string template https://github.com/smarty-php/smarty/issues/639
+
 ## [3.1.39] - 2021-02-17
 
 ### Security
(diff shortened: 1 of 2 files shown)


wisskid:
@AnrDaemon have a look please?

AnrDaemon:
Had one. There's at least two issues with this PR.

1. It evaluates incorrectly (see modified unit test in #652 ).
2. It does not cover the regression reported (see specific regression test in the same PR).

wisskid:
so... is the fuss about the newline or about the fatal error?

AnrDaemon:
Now it is about different rendering result before and after the fix.

There's no more crashes as far as I can see.

wisskid:
OK, but In my experience there was *always* a fatal error when rendering a string template inline. 
I see the extra newline in the output, but wouldn't that be a separate, somewhat less important issue? Also: why would one render a string template as inline?

AnrDaemon:
`string:` source was used to easily demonstrate the issue using as few extra files as possible.
In real world, it was indeed an external file, but I made sure the results are exactly same.

AnrDaemon:
To clarify: I've never had crashes with `string:… inline` alone.

PR #665, opened 2021-05-27
Modify links from http to https in composer.json

matks:
This is definitely not a huge contribution, but this looks a little better and can reassure some persons ^^

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
     "keywords": [
         "templating"
     ],
-    "homepage": "http://www.smarty.net",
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
I wonder if that freenode link should be updated? I assume smarty isn't Freenode after all the latest drama.

matks:
Waoh, looks like I keep being myself 😅 I already submitted similar PR https://github.com/smarty-php/smarty/pull/643/ before. Sorry for the spam. When I see http links in my composer.lock I immediately submit PRs to upstream repos.

PR #667, opened 2021-05-28
Append target file ext/name to tmp file (speedup antivirus scans)

mvorisek:
No file extension causes issues with Microsoft Antivirus, this simple update increase performance about 5x.

Was super hard to debug, but once ".php" suffix is added, performance is fixed consistently on Windows.

A bugfix release should be done once merged.

### DIFF
--- a/libs/sysplugins/smarty_internal_runtime_writefile.php
+++ b/libs/sysplugins/smarty_internal_runtime_writefile.php
@@ -54,7 +54,10 @@ class Smarty_Internal_Runtime_WriteFile
             }
         }
         // write to tmp file, then move to overt file lock race condition
-        $_tmp_file = $_dirpath . DIRECTORY_SEPARATOR . str_replace(array('.', ','), '_', uniqid('wrt', true));
+        // use the original file extension/basename to prevent extensive file scans
+        // by antiviruses like Microsoft Defender
+        $_tmp_file = $_dirpath . DIRECTORY_SEPARATOR . '.' . str_replace(array('.', ','), '_', uniqid('wrt', true))
+            . '.' . basename($_filepath);
         if (!file_put_contents($_tmp_file, $_contents)) {
             error_reporting($_error_reporting);
             throw new SmartyException("unable to write file {$_tmp_file}");


wisskid:
@mvorisek  Smarty_Internal_Runtime_WriteFile is a general purpose class, not specifically designated for php files. Would a '.tmp' extension work too?

mvorisek:
> @mvorisek Smarty_Internal_Runtime_WriteFile is a general purpose class, not specifically designated for php files. Would a '.tmp' extension work too?

Makes sense. I did some experiments, and yes, if explicitly defined as excluded in https://docs.microsoft.com/en-us/microsoft-365/security/defender-endpoint/configure-extension-file-exclusions-microsoft-defender-antivirus basically any extension can be used.

It is however impractical. So I changed it to append the whole basename. Whole basename instead of basename extension only as the exclusion rules may be defined for some longer file part than the the part after the last dot in general.

Any concerns about this approach?

mvorisek:
@wisskid can you please merge and tag a new release?

mvorisek:
can this fix be merged please?

mvorisek:
@wisskid can you please merge?

wisskid:
@mvorisek I'm not sure this is a good idea. Nothing is broken, it's not a bug and I'm not 100% sure this wouldn't break anything or cause issues for someone else. Another antivirus application might consider to do an indepth scan "your" filenames, for example. Since this is basically appears to be a Microsoft Antivirus issue, it would seem to better to fix it there.

mvorisek:
@wisskid can you please reconsider? Additionally, naming the files with php ext will present the file sources to be downloadable.

mvorisek:
@wisskid can you please kind reopen this PR? The speedup on Windows is really dramatic and isn't is actually better to sue the original filename extension?

wisskid:
@mvorisek I've reconsidered and still don't want to change this just because some Antivirus application doesn't get it. I read the page at [your link](https://docs.microsoft.com/en-us/microsoft-365/security/defender-endpoint/configure-extension-file-exclusions-microsoft-defender-antivirus) and I noticed you can easily exclude a directory. I propose you configure the cache and compile dirs to be excluded from scanning.
You also mention "Additionally, naming the files with php ext will present the file sources to be downloadable.". You probably mean "prevent". In any case: please configure your web application to not store cached or compiled templates under the web root. They should never be downloadable or runnable.

mvorisek:
@wisskid The problem is degraded performace very, very hard to identify. Many users do not have a clue the slowdown is caused by antivirus, thus they do not/cannot take any action/add an exclude antivirus rule. And it is never good to advise to disable antivirus, it does sound insecure and some company politics even forbids that explicitly. So please reconsider this PR vs. the disadvantages, they is no major disadvantage againts.

mvorisek:
@wisskid can this PR be reopened and merged, please?

Issue #671, opened 2021-07-26
Smarty is not compatible with PHP 8.1


ildar-aim:
https://github.com/php/php-src/blob/php-8.1.0/UPGRADING

https://www.php.net/index.php#id2021-07-22-1

For example, this patch causes a fatal error

```
aim-server[~/www/smarty/demo]$ gitdiff .
diff --git a/demo/templates/index.tpl b/demo/templates/index.tpl
index 1fbb6d37..1e125e15 100644
--- a/demo/templates/index.tpl
+++ b/demo/templates/index.tpl
@@ -1,6 +1,6 @@
 {config_load file="test.conf" section="setup"}
 {include file="header.tpl" title=foo}
-
+{php}{/php}
 <PRE>
 
 {* bold and title are read from the config file *}
```

```
Fatal error: Type of SmartyCompilerException::$line must be int (as in class Exception) in public_html/smarty/libs/sysplugins/smartycompilerexception.php on line 8
```

The expected error is 
```
Fatal error: Uncaught --> Smarty Compiler: Syntax error in template "file:public_html/smarty/demo/templates/index.tpl" on line 3 "{php}{/php}" {php}{/php} tags not allowed. Use SmartyBC to enable them <-- thrown in public_html/smarty/libs/sysplugins/smarty_internal_templatecompilerbase.php on line 3
```

pilif:
what's really nasty is that fixing the class to work with 8.1 makes it break for < 8.1.

the only fix I see is to either conditionally define the class depending on PHP version or alternatively renaming the property and using the changed name instead.

Which one would you prefer? I can do a PR for both.

wisskid:
See https://github.com/smarty-php/smarty/pull/678

genyslt:
Any news?

glensc:
if the problem being `$line` being re-declared as `public` from inherited `Exception` class, can it be declared as private if used only in that class?

wisskid:
@glensc yes, see #678

PR #678, opened 2021-10-07
Add PHP 8.1 compatibility

atomiix:
The goal of this PR is to make smarty compatible with PHP 8.1.

The main issue was the use of the function `strftime()` which is deprecated in PHP 8.1. This problem is solve by the use of `IntlDateFormatter` after being sure that the `intl` extension is loaded.

Other deprecations have been removed too.

### DIFF
--- a/libs/plugins/function.html_select_date.php
+++ b/libs/plugins/function.html_select_date.php
@@ -306,11 +306,36 @@ function smarty_function_html_select_date($params, Smarty_Internal_Template $tem
             $_html_months .= '<option value="">' . (isset($month_empty) ? $month_empty : $all_empty) . '</option>' .
                              $option_separator;
         }
+
+        $formatter = null;
+        $format_compare = '%m';
+        if (class_exists('IntlDateFormatter')) {
+            $format_compare = 'm';
+            $patterns = array('%b', '%h', '%B', '%m');
+            $replacement = array('MMM', 'MMM', 'MMMM', 'MM');
+            $month_format = str_replace($patterns, $replacement, $month_format);
+            $month_value_format = str_replace($patterns, $replacement, $month_value_format);
+            $formatter = new IntlDateFormatter(
+                setlocale(LC_TIME, '0'),
+                IntlDateFormatter::NONE,
+                IntlDateFormatter::NONE)
+            ;
+        }
         for ($i = 1; $i <= 12; $i++) {
+            if (null !== $formatter) {
+                $formatter->setPattern($month_format);
+                $_text = $formatter->format($_month_timestamps[ $i ]);
+                $formatter->setPattern($month_value_format);
+                $_value = $formatter->format($_month_timestamps[ $i ]);
+            } else {
+                $_text = strftime($month_format, $_month_timestamps[ $i ]);
+                $_value = strftime($month_value_format, $_month_timestamps[ $i ]);
+            }
+
             $_val = sprintf('%02d', $i);
             $_text = isset($month_names) ? smarty_function_escape_special_chars($month_names[ $i ]) :
-                ($month_format === '%m' ? $_val : strftime($month_format, $_month_timestamps[ $i ]));
-            $_value = $month_value_format === '%m' ? $_val : strftime($month_value_format, $_month_timestamps[ $i ]);
+                ($month_format === $format_compare ? $_val : $_text);
+            $_value = $month_value_format === $format_compare ? $_val : $_value;
             $_html_months .= '<option value="' . $_value . '"' . ($_val == $_month ? ' selected="selected"' : '') .
                              '>' . $_text . '</option>' . $option_separator;
         }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,7 @@ jobs:
           - "7.3"
           - "7.4"
           - "8.0"
+          - "8.1"
 
         compiler:
           - default
--- a/libs/Smarty.class.php
+++ b/libs/Smarty.class.php
@@ -169,6 +169,11 @@ class Smarty extends Smarty_Internal_TemplateBase
     public static $_DATE_FORMAT = '%b %e, %Y';
 
     /**
+     * The date format to be used internally by IntlDateFormatter
+     */
+    public static $_ICU_DATE_FORMAT = 'MMM d, yyyy';
+
+    /**
      * Flag denoting if PCRE should run in UTF-8 mode
      */
     public static $_UTF8_MODIFIER = 'u';
--- a/libs/plugins/modifier.date_format_intl.php
+++ b/libs/plugins/modifier.date_format_intl.php
@@ -1,0 +1,62 @@
+<?php
+/**
+ * Smarty plugin
+ *
+ * @package    Smarty
+ * @subpackage PluginsModifier
+ */
+/**
+ * Smarty date_format_intl modifier plugin
+ * Type:     modifier
+ * Name:     date_format_intl
+ * Purpose:  format datestamps via IntlDateFormatter
+ * Input:
+ *          - string: input date string
+ *          - format: ICU format for output
+ *          - default_date: default date if $string is empty
+ *
+ * @link   https://www.smarty.net/manual/en/language.modifier.date.format.php date_format (Smarty online manual)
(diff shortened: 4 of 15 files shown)


atomiix:
@wisskid just a friendly reminder as PHP 8.1 due date is in 2 days 🙃

Progi1984:
Good news. PHP 8.1 is available ;)

libs/sysplugins/smarty_internal_method_loadplugin.php
```diff
@@ -84,7 +84,7 @@ public function loadPlugin(Smarty $smarty, $plugin_name, $check)
             // loop through plugin dirs and find the plugin
             foreach ($_p_dirs as $_plugin_dir) {
                 foreach ($_file_names as $name) {
-                    $file = $_plugin_dir . $name;
+                    $file = $_plugin_dir . strtolower($name);
```
wisskid, line 87: The code surrounding this change has explicit checks for lowercasing. This change would break when loading a plugin with a case sensitive name on case sensitive filesystems.

libs/plugins/function.html_select_date.php
```diff
@@ -306,11 +306,36 @@ function smarty_function_html_select_date($params, Smarty_Internal_Template $tem
             $_html_months .= '<option value="">' . (isset($month_empty) ? $month_empty : $all_empty) . '</option>' .
                              $option_separator;
         }
+
+        $formatter = null;
+        $format_compare = '%m';
+        if (class_exists('IntlDateFormatter')) {
```
wisskid, line 312: Since `IntlDateFormatter` isn't included by default and the `strftime` deprecation notice is only a notice, wouldn't it be better to keep it as it was? We're adding complexity here, only fixing the deprecation notices on certain environments. I'm hoping PHP9 will include a way to handle this smoothly that is included by default?

wisskid (changes requested):
Great PR by the way, 2 comments made

libs/plugins/function.html_select_date.php
atomiix: Unfortunately, I don't think that PHP 9.0 will handle it smoothly, from [https://wiki.php.net/rfc/deprecations_php_8_1](https://wiki.php.net/rfc/deprecations_php_8_1):
>The strftime() and gmstrftime() functions exhibit similar issues as strptime(), in that the formats they support, as well as their behavior, is platform-dependent. Unlike strptime(), these functions are available on Windows, though with a different feature set than on Linux. Musl-based distributions like Alpine do not support timezone-related format specifiers correctly. These functions are also locale-based, and as such may exhibit thread-safety issues.

> Once again date() or DateTime::format() provide portable alternatives, and IntlDateFormatter::format() provides a more sophisticated, localization-aware alternative.

> The proposal is to deprecate strftime() and gmstrftime() in favor of these alternatives.

So I guess this is the way we'll have to do it in PHP 9.0 and in the meantime, we can have `strftime` as a fallback. WDYT?

wisskid: Let me put it in another way: your change will still fail in PHP9, because `strftime` will be deleted by then (unless they reverse the decision). So what is the point of making the change in the first place?
Also: I know `strftime` is used a LOT in PHP libs, so the deprecation will trigger the need for a suitable alternative to be included in PHP9 by default. If that turns out to be `IntlDateFormatter`, we'll switch then.

PierreRambaud: I agree with @atomiix , having `strftime` as fallback is a good idea. I'm not in favor of having deprecation notices only because it's acceptable. When you're working you don't want to be spamed by useless messages you can avoid.
Adding `IntlDateFormatter` as alternative is a good idea, but the real check should be `if (extension_loaded('intl')) {` for a better understanding.

iquito: I also think avoiding any `strftime` would be beneficial. The intl extension is quite common nowadays, and Symfony has a polyfill (`symfony/polyfill-intl-icu`) which is an easy alternative if the intl extension is missing and provides a way of avoiding deprecation notices even without the extension.

pineapps:
So are we going to merge this branch or.....

PierreRambaud:
> So are we going to merge this branch or.....?

Maybe, maybe not, that's a hard question :thinking: 
Let the maintainer take time to estimate if the change is legit and can be accepted or not :smiley_cat:

j-applese3d:
I noticed that [`modifier.date_format`](https://github.com/smarty-php/smarty/blob/cd962280ce9af71e35d64248f85d4fe4b6cf7db5/libs/plugins/modifier.date_format.php#L81) also uses `strftime`.
Would it make sense to also update that plugin at the same time?

PierreRambaud:
> I noticed that [`modifier.date_format`](https://github.com/smarty-php/smarty/blob/cd962280ce9af71e35d64248f85d4fe4b6cf7db5/libs/plugins/modifier.date_format.php#L81) also uses `strftime`. Would it make sense to also update that plugin at the same time?

I think yes, but the whole part needs to be refactored: https://github.com/smarty-php/smarty/blob/cd962280ce9af71e35d64248f85d4fe4b6cf7db5/libs/plugins/modifier.date_format.php#L51

atomiix:
@j-applese3d you're right! I just added a new modifier [`date_format_intl`](https://github.com/smarty-php/smarty/pull/678/files#diff-adb0561d32e256a7dce97f0192179d02bfb8d0567169cf6bdd216d3c43c2f910). I guess that's a good compromise to avoid a BC break and still allow us to use `IntlDateFormatter` when needed.

PierreRambaud:
> @j-applese3d you're right! I just added a new modifier [`date_format_intl`](https://github.com/smarty-php/smarty/pull/678/files#diff-adb0561d32e256a7dce97f0192179d02bfb8d0567169cf6bdd216d3c43c2f910). I guess that's a good compromise to avoid a BC break and still allow us to use `IntlDateFormatter` when needed.

Not sure it's the right solution as the `date_format` will still be used.
Maybe you can do the same condition inside the current implementation as you did before, like "Is Intl available, yes, let's use it". And use it by default.

atomiix:
>Maybe you can do the same condition inside the current implementation as you did before, like "Is Intl available, yes, let's use it". And use it by default.

The issue is that `IntlDateFormatter` cannot be 100% compatible with `strftime`. You can see in [this gist](https://gist.github.com/ve3/a3b7924a85c3286b554f9e636b919883) that it's not 100% correct.

IMHO, for `date_format`, as you can fully customize the format, it's better to have a dedicated method

wisskid:
I like the idea of combining `IntlDateFormatter` with `symfony/polyfill-intl-icu` and replace all the `strftime` in the entire codebase.

atomiix:
@wisskid We'll have to keep in mind that `symfony/polyfill-intl-icu` is limited to english though. That being said, that's ok for me. Also, that means BC breaks as `strftime` format string is different than `IntlDateFormatter`'s ones and not completely convertisable.

iwiznia:
I am working on a migration of a project to PHP 8.1 and can confirm that this PR solves all the problems I've been seeing.

genyslt:
Any news?

wisskid:
I'm thinking we should separate the fixes for the fatal error on PHP8.1(#671) from the removal of deprecation notices strftime. The former is much more pressing now.

genyslt:
Can you tell us estimated time, when you can fix it? Is there a lot of problems to run Smarty on php8.1?

thirsch:
Looks like there is not much missing to get 8.1 support. Is there anything we can do to speed up the merge/release? Maybe creating another PR with all changes except the strftime-issue?

wisskid:
> Maybe creating another PR with all changes except the strftime-issue?

Yes that would be super. I'd merge that.

thirsch:
> > Maybe creating another PR with all changes except the strftime-issue?
> 
> Yes that would be super. I'd merge that.

Okay, I've created another pr for only the non-intldate related stuff: https://github.com/smarty-php/smarty/pull/711

scortellino:
There are 3 PR on PHP 8.1...Do we expect a merge soon like in coming days?

wisskid:
Closing this one in favor of https://github.com/smarty-php/smarty/pull/713
We'll handle the strftime issue later

PR #687, opened 2021-12-01
Fix for php 8.0

pandarek:
$parts must be an array, because the argument for count() must be countable, mb_split() returns an array or a boolean.

### DIFF
--- a/libs/plugins/shared.mb_str_replace.php
+++ b/libs/plugins/shared.mb_str_replace.php
@@ -44,9 +44,9 @@ if (!function_exists('smarty_mb_str_replace')) {
                 }
             }
         } else {
-            $parts = mb_split(preg_quote($search), $subject);
-            $parts = $parts ?: array(0 => '');
-            $count = is_countable($parts) ? count($parts) - 1 : 0;
+            $parts = mb_split(preg_quote($search, null), $subject);
+            $parts = $parts ?: array();
+            $count = count($parts) - 1;
             $subject = implode($replace, $parts);
         }
         return $subject;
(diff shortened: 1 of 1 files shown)


libs/plugins/shared.mb_str_replace.php
```diff
@@ -44,7 +44,8 @@ function smarty_mb_str_replace($search, $replace, $subject, &$count = 0)
                 }
             }
         } else {
-            $parts = mb_split(preg_quote($search), $subject);
+            $parts = mb_split(preg_quote($search, null), $subject);
```
wisskid, line 47: `, null` seems unnecessary?

libs/plugins/shared.mb_str_replace.php
```diff
@@ -44,7 +44,8 @@ function smarty_mb_str_replace($search, $replace, $subject, &$count = 0)
...
+            $parts = $parts ?: array();
```
wisskid, line 48: you could add `?: array()` to the end of the previous line as well

libs/plugins/shared.mb_str_replace.php
pandarek: Correct, it can be omitted

libs/plugins/shared.mb_str_replace.php
pandarek: yes, it will be nicer

pandarek:
corrections made

wisskid (approved):

wisskid:
@pandarek  thank you!

PR #704, opened 2022-01-11
Update project for PHP8.1

Indianos:
Since PHP8.1 is asking strict type, removed `$line = null` default value in smartycompilerexception.php

### DIFF
--- a/libs/sysplugins/smartycompilerexception.php
+++ b/libs/sysplugins/smartycompilerexception.php
@@ -18,9 +18,9 @@ class SmartyCompilerException extends SmartyException
     /**
      * The line number of the template error
      *
-     * @type int|null
+     * @type int
      */
-    public $line = null;
+    public $line;
 
     /**
      * The template source snippet relating to the error


libs/sysplugins/smartycompilerexception.php
```diff
@@ -18,9 +18,9 @@ public function __toString()
     /**
      * The line number of the template error
      *
-     * @type int|null
+     * @type int
      */
-    public $line = null;
+    public $line;
```
thirsch, line 23: Unfortunately, this does not solve the issue for us. We are still getting the following error:

```NOTICE: PHP message: PHP Fatal error:  Type of SmartyCompilerException::$line must be int (as in class Exception) in /var/ftproot/www/symfony/lib/vendor/smarty/smarty/libs/sysplugins/smartycompilerexception.php on line 8```

If I change the line to `public int $line;`, everything is working, but this is not backward compatible to php < 7.4

wisskid:
Please see https://github.com/smarty-php/smarty/pull/713

Issue #199, opened 2016-03-02
Strange processing of built-in functions 3.1.27 vs 3.1.29


Stevad:
I found some strange behaviour with Smarty built-in functions. For examle - `mailto`.

If I add `{mailto address='mail@mail.ua'}` somewhere in template then:

- v3.1.27 - works as expected
- v3.1.28, v3.1.29 before displaying result of the function will try to load class `Smarty_Internal_Compile_Mailto` and generates PHP warnings like these

`
Warning: include(Smarty_Internal_Compile_Mailto.php): failed to open stream: No such file or directory in /some/path/to/my/project
`

Where to look for errors? 
P.S. I am using Smarty with Yii 1.x. Smarty installed via composer.

uwetews:
I have no idea how the autolader could run into this warning. But I reverted some autoloader changes.

The fix is in the master branch an will later be included in 3.1.30

KarelWintersky:
Same issue with 3.1.33

PR #713, opened 2022-01-18
PHP 8.1 compatibility (without IntlDateFormatter)

thirsch:
This is the follow-up for https://github.com/smarty-php/smarty/pull/711. I tried to keep as many commits without modification here. Hope that's ok @GoT

I coudn't find a way to update the previous pr, sorry.

### DIFF
--- a/libs/sysplugins/smarty_internal_cacheresource_file.php
+++ b/libs/sysplugins/smarty_internal_cacheresource_file.php
@@ -196,8 +196,8 @@ class Smarty_Internal_CacheResource_File extends Smarty_CacheResource
      */
     public function hasLock(Smarty $smarty, Smarty_Template_Cached $cached)
     {
-        clearstatcache(true, $cached->lock_id);
-        if (is_file($cached->lock_id)) {
+        clearstatcache(true, $cached->lock_id ?? '');
+        if (null !== $cached->lock_id && is_file($cached->lock_id)) {
             $t = filemtime($cached->lock_id);
             return $t && (time() - $t < $smarty->locking_timeout);
         } else {
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,7 @@ jobs:
           - "7.3"
           - "7.4"
           - "8.0"
+          - "8.1"
 
         compiler:
           - default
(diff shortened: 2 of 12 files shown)


libs/sysplugins/smarty_internal_cacheresource_file.php
```diff
@@ -196,8 +196,8 @@ public function clear(Smarty $smarty, $resource_name, $cache_id, $compile_id, $e
      */
     public function hasLock(Smarty $smarty, Smarty_Template_Cached $cached)
     {
-        clearstatcache(true, $cached->lock_id);
-        if (is_file($cached->lock_id)) {
+        clearstatcache(true, $cached->lock_id ?? '');
+        if (null !== $cached->lock_id && is_file($cached->lock_id)) {
```
glensc, line 200: since you closed previous one, carry the discussion:
- https://github.com/smarty-php/smarty/pull/711#discussion_r786668634

thirsch: Shall we continue the discussion here? My last answer was:

> Personally I agree with your point, but IMO it's out of scope of this change. What the original author tried to do is "doing whatever php < 8.1 would do". So another pr should cope with the issue.

glensc: I have no say of this, as I'm not the maintainer.

personally, I would carry the fix to separate PR and hope it gets merged sooner as the change is smaller.

ps: edit `#199` to be backticks as it links to bogus ticket, or replace with link to source

trasher:
Hi,
There are still `strftime` calls that will make Smarty fail with PHP 8.1:
```
libs/plugins/function.html_select_date.php:                ($month_format === '%m' ? $_val : strftime($month_format, $_month_timestamps[ $i ]));
libs/plugins/function.html_select_date.php:            $_value = $month_value_format === '%m' ? $_val : strftime($month_value_format, $_month_timestamps[ $i ]);
libs/plugins/modifier.date_format.php:        return strftime($format, $timestamp);
```

Can a maintainer approve the actions workflow run so we can also see tests results? Thanks!

thirsch:
> Hi, There are still `strftime` calls that will make Smarty fail with PHP 8.1 [...]

yes, please see the discussion in #678. This pr is just a follow-up without fixing the strftime issues. These are deprecations only and the main goal is to get smarty running under 8.1.

trasher:
Yes, I've realized. Thanks.

iwiznia:
Nice, thanks @thirsch!! 
@wisskid is there going to be a new version release to include these changes?

thirsch:
Kudos clearly must go to @atomiix!

acun87 (commented):
id_ed25519.pub

PR #714, opened 2022-01-20
Update smarty_internal_runtime_codeframe.php

avatec:
Update for php 8.1 strftime is deprecated


artcs:
You cannot simply replace `strftime` with `date`. You have to update the formatstring accordingly (e.g. `'Y-m-d H:i:s'` instead of `%Y-%m-%d %H:%M:%S`).

avatec:
> You cannot simply replace `strftime` with `date`. You have to update the formatstring accordingly (e.g. `'Y-m-d H:i:s'` instead of `%Y-%m-%d %H:%M:%S`).

You are right - so only need to remove those % and it will be done.

artcs:
> so only need to remove those % and it will be done.

Not exactly. E.g. `%M` has to be replaces by `i`. Read the manual.

wisskid:
@avatec we will work on a more comprehensive replacement for strftime soon, covering all occurences.

PR #716, opened 2022-01-20
Update modifier.truncate.php

avatec:
For PHP 8.1 if string is empty. Repairs Deprecated: mb_strlen(): Passing null to parameter #1 ($string) of type string is deprecated in vendor/smarty/smarty/libs/plugins/modifier.truncate.php on line 33

### DIFF
--- a/libs/plugins/modifier.truncate.php
+++ b/libs/plugins/modifier.truncate.php
@@ -29,6 +29,9 @@ function smarty_modifier_truncate($string, $length = 80, $etc = '...', $break_wo
     if ($length === 0) {
         return '';
     }
+    if(empty( $string )) {
+        return;
+    }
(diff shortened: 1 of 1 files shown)


Visualq:
This is BC breaking, considering `empty()` returns true on "0","false", "", [] and null and your return type is "null" hence the `return; `

It would be better to:

`if ($string === null) {
    return ''; 
}`

 However, the null problem might require a more general fix.

avatec:
Yes, but i think no one should truncate 0, false or empty array. It should be checked before. We can also use !isset instead of empty.

wisskid:
Agree with @Visualq here. Also, we're going to needs tons of (string) casts or null checks to weed out every deprecation warning, I guess.

PR #718, opened 2022-01-21
Revert regression: no error on array key access

raveren:
Makes it so that not only `null` but `false` variables don't throw errors when accessed as array - as it was with Smarty 3.*


wisskid:
this is fixed in #830

Issue #706, opened 2022-01-12
PHP 8.1 deprecation notices in demo/plugins/cacheresource.pdo.php


alexgit2k:
With PHP 8.1 there are several deprecation notices in demo/plugins/cacheresource.pdo.php:
```
line 247: Optional parameter $cache_id declared before required parameter $content is implicitly treated as a required parameter
line 199: Optional parameter $cache_id declared before required parameter $mtime is implicitly treated as a required parameter
line 247: Optional parameter $compile_id declared before required parameter $content is implicitly treated as a required parameter
line 199: Optional parameter $compile_id declared before required parameter $mtime is implicitly treated as a required parameter
```

marclaporte:
Related: https://github.com/smarty-php/smarty/pull/678

PR #720, opened 2022-02-04
cacheresource.pdo.php deprecation warning fix

milanstojanovic88:
I've been having deprecation warning due to wrong order of optional arguments in these two methods. This is just a suggestion fix.


wisskid:
I don't think this fixes anything. In case I'm missing something, please submit a proper bug report.

wisskid:
@milanstojanovic88 I just saw #706, fixing that now, slightly differently.

PR #730, opened 2022-02-15
namespace-based plugins using autoload()

SlowFox71:


wisskid:
This looks promising. Could you provide an explanation of what you are trying to accomplish?

SlowFox71:
Personally, main motivation was the use of namespace-based constants in plugins, i.e. being able to write things like

```<?php

namespace SmartyPlugins;

class Link {

        const DECORATION_NONE = 'none';
        const DECORATION_BUTTON = 'button';
        const DECORATION_ICON = 'icon';

        public static function block($params, $content, $smartyInternalTemplate, &$repeat) {
           [...do the magic...]
        }
}
```
and being able to use this in my template as

```{link page=\Page\SomePage::class decoration=\SmartyPlugins\Link::DECORATION_BUTTON}Text{/link}```

Other than with magic strings I can (via AST) automatically check templates for correct/existing constants which (for me) was a huge improvement for the quality of user-modified templates.

Another benefit is auto-loading via PHP-autoloader (instead of doubling this within Smarty). This allows me to intercept plugin-loading at the same point where I do it for all the other classes.

Possible future benefit could be implementation of block/function within the same class, e.g.
```
{link page=\Page\PageProvidingOwnLinkText::class} 
{link page=\Page\PageProvidingOwnLinkText::class}Customized link text{/link}
```
However, this would require modifications in the parser.

For a distant future it could be a step to the general usage of namespaces by Smarty. Internal plugins could use this logic (and then easily being overloaded via Autoloader if necessary).

PR #735, opened 2022-02-18
Fix error handler allowUndefinedVars

rudiedirkx:
`handleError` only checks for property `value`, which is odd. I'm getting error

    Attempt to read property "id" on null

Because `id` is the property being accessed. So it should check for any property name, not just `value`.


libs/sysplugins/smarty_internal_errorhandler.php
```diff
@@ -66,7 +66,7 @@ public function deactivate() {
      */
     public function handleError($errno, $errstr, $errfile, $errline, $errcontext = [])
     {
-        if ($this->allowUndefinedVars && $errstr == 'Attempt to read property "value" on null') {
+        if ($this->allowUndefinedVars && preg_match('/^Attempt to read property ".+?" on null$/', $errstr)) {
```
sapics, line 69: I want to use `\w` instead of `.`, when we need to catch alphanumeric case.

rudiedirkx: Properties can also include `-` and maybe even `:` etc, so it **uneagerly** matches to the first `"`. https://3v4l.org/s9KPc

sapics: It makes sense.

Martijn-Bos:
This pull request would make smarty v4.1 + php 8.1 useable. Bump

rudiedirkx:
Does nobody have `allowUndefinedVars` enabled in PHP 8.0+?

wisskid:
This change seems straightforward. Should be able to merge this. Just to be clear: could you provide an exact issue? I.e. what problem does this solve?

rudiedirkx:
```
$smarty->assign('object', null);
====
{{$object->name}}
```

That would trigger a

> Attempt to read property "name" on null

----

I'll fix the merge conflict.

wisskid:
I propose a slightly different implementation, see https://github.com/smarty-php/smarty/pull/832

Issue #672, opened 2021-07-27
strftime deprecated in PHP 8.1


mambax7:
```
Unknown: Function strftime() is deprecated in file /Smarty_Compiler.class.php line 395
Unknown: preg_match_all(): Passing null to parameter #2 ($subject) of type string is deprecated in file /Smarty_Compiler.class.php line 1531
```

Bigbear7:
The strftime deprecation was actually a bit controversial where the RFC author voted against it ;-) Also the benefit of deprecating  was debated on the internals list, see: https://externals.io/message/113657

Anyway, is this also an issue for Smarty 3?

mambax7:
1) Nikita was not the author of it, he just collected the various deprecation requests:
> "This is a collection of minor deprecations that various people have put together over the last ~2 years."

2) I don't know about Smarty 3 - we're still using Smarty 2

Bigbear7:
Well, when checking for the prevalence of strftime in Smarty 3 I found it in the following files:
- /plugins/function.html_select_date.php
- /plugins/modifier.date_format.php
- /Smarty.class.php
- /sysplugins/smarty_internal_config_file_compiler.php
- /sysplugins/smarty_internal_runtime_codeframe.php

Probably the same for Smarty 2.

wisskid:
Thank you for your report. We'll fix this in due time. But let's get a PHP8.0 release out first :)

mambax7:
@wisskid Thank you! 
Now that PHP 8.1 is available, I hope that this update will be available soon too! :)

mambax7:
I see that 4.x is moving very quickly with PHP 8.1 updates. 
When could we expect  an update for Smarty 2.x?
Pretty Please? :)

wisskid:
@mambax7 I have no such intentions, but feel free to provide a PR!

mambax7:
So is Smarty 2.x now officially "**Abandonware**"?

wisskid:
@mambax7 no, Smarty has a v3 and a v4 you could upgrade to.

mambax7:
I understand that, but is Smarty **2.x** now officially **"Abandonware"**?

wisskid:
[...]

PR #745, opened 2022-03-14
Support PHP 8.1 for version 2.6: Fix deprecated warning when calling strftime for template compiling

Atanamo:
Fixes deprecation of strftime() since PHP 8.1 in branch support/2.6.

### DIFF
--- a/libs/Smarty_Compiler.class.php
+++ b/libs/Smarty_Compiler.class.php
@@ -391,7 +391,7 @@ class Smarty_Compiler extends Smarty {
         }
 
         // put header at the top of the compiled template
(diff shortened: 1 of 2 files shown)


Mike-Benoit:
A similar change to this will also fix Smarty v3.x and it only affects comments in compiled templates, so its extremely harmless. Any chance on getting this merged soon?

wisskid:
Smarty v2 is not supported anymore. If you want to run PHP8.0+, please upgrade to Smarty v4.
The deprecation warnings will be handled soon, but there are many more instances to be fixed and work is already underway. Please see https://github.com/smarty-php/smarty/issues/750 en https://github.com/smarty-php/smarty/issues/672

Issue #750, opened 2022-03-30
PHP 8.1: Deprecation notices for several functions in Smarty 3 and 4


alexgit2k:
There are several deprecation-notices with PHP 8.1 in Smarty 3 and 4:

- `explode`: Passing null to parameter
- `number_format`: Passing null to parameter
- `escape`: Passing null to parameter (see #748)
- `date_format`:  strftime() is deprecated (see #672)
- `replace`: Passing null to parameter

### Example

```
docker run -it --rm php:8.1 /bin/bash

apt-get update -yqq && apt-get install -y git
curl https://getcomposer.org/download/latest-stable/composer.phar > /usr/local/bin/composer
chmod a+rx /usr/local/bin/composer
composer require smarty/smarty:^3

cat <<EOT > test.php
<?php
require_once('vendor/autoload.php');
\$smarty = new Smarty();
\$smarty->assign('value',null);

print '- {\$value|escape} -> htmlspecialchars(\$value, ENT_QUOTES, \'UTF-8\', true);';
\$smarty->display('string:{\$value|escape}');

print '- {count(";"|explode:\$value)} -> explode(";",\$value);';
\$smarty->display('string:{count(";"|explode:\$value)}');

print '- {\$value|number_format} -> number_format(\$value);';
\$smarty->display('string:{\$value|number_format}');

print '- {\$smarty.now|date_format} -> strftime(\$format, \$timestamp);';
\$smarty->display('string:{\$smarty.now|date_format}');

print '- {\$value|replace:"":""} -> mb_split(preg_quote(\$search), \$subject);';
\$smarty->display('string:{\$value|replace:"":""}');

?>
EOT

php test.php
```

Same deprecation notices with Smarty 4: `composer require smarty/smarty`
No deprecation notices with PHP 8.0: `docker run -it --rm php:8.0 /bin/bash`

### Output

- {$value|escape} → htmlspecialchars($value, ENT_QUOTES, 'UTF-8', true);
`Deprecated: htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated in compiled template on line 18`

- {count(";"|explode:$value)} → explode(";",$value);
`Deprecated: explode(): Passing null to parameter #2 ($string) of type string is deprecated in compiled template on line 18`

- {$value|number_format} → number_format($value);
`Deprecated: number_format(): Passing null to parameter #1 ($num) of type float is deprecated in compiled template on line 18`

[...]

PR #755, opened 2022-05-03
address PHP 8.1 'explode', 'number_format', and 'replace' deprecations

mfettig:
This PR is meant to address several of the PHP 8.1 deprecation items noted in Issue #750 

It does not address the deprecation noted in 'escape' because I was not able to recreate the issue against the current master branch

It also does not address any strftime() related deprecation issues since it is not clear to me how that issue should be addressed as a whole. Discussions were present in #678 and touched on using an intl polyfill, but changes were deferred at that time.  #672 is open as a strftime() specific issue so I suppose it would still be handled there unless a decision has already been made, in which case I would be happy to update this PR once I know how it should be handled.

### DIFF
--- a/libs/plugins/modifier.number_format.php
+++ b/libs/plugins/modifier.number_format.php
@@ -1,0 +1,26 @@
+<?php
+/**
+ * Smarty plugin
+ *
+ * @package    Smarty
+ * @subpackage PluginsModifier
+ */
+
+/**
+ * Smarty number_format modifier plugin
+ * Type:     modifier
+ * Name:     number_format
+ * Purpose:  Format a number with grouped thousands
+ *
+ * @param float|null  $num
+ * @param int         $decimals
+ * @param string|null $decimal_separator
+ * @param string|null $thousands_separator
+ *
+ * @return string
+ */
+function smarty_modifier_number_format(?float $num, int $decimals = 0, ?string $decimal_separator = ".", ?string $thousands_separator = ",")
+{
+    // provide $num default to prevent deprecation errors in PHP >=8.1
+    return number_format($num ?? 0.0, $decimals, $decimal_separator, $thousands_separator);
+}
--- a/tests/UnitTests/TemplateSource/TagTests/PluginModifier/PluginModifierExplodeTest.php
+++ b/tests/UnitTests/TemplateSource/TagTests/PluginModifier/PluginModifierExplodeTest.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace UnitTests\TemplateSource\TagTests\PluginModifier;
+
+/**
+ * class for modifier tests
+ *
+ * @runTestsInSeparateProcess
+ * @preserveGlobalState disabled
+ * @backupStaticAttributes enabled
+ */
+class PluginModifierExplodeTest extends \PHPUnit_Smarty
+{
+    public function setUp(): void
+    {
+        $this->setUpSmarty(__DIR__);
+    }
+
+    /**
+     * @return void
+     * @throws \SmartyException
+     *
+     * @dataProvider explodeDataProvider
+     */
+    public function testExplode($template, $subject, $expectedString)
+    {
+        $this->smarty->assign('subject', $subject);
+
+        $tpl = $this->smarty->createTemplate($template);
+        $res = $this->smarty->fetch($tpl);
+
+        $this->assertEquals($expectedString, $res);
+    }
+
+    public function explodeDataProvider()
+    {
+        return [
+            'default'  => [
+                'template'      => 'string:{","|explode:$subject|json_encode}',
+                'subject'       => 'a,b,c,d',
+                'expectedString' => '["a","b","c","d"]',
+            ],
+            'withNoDelimiterFound'  => [
+                'template'      => 'string:{","|explode:$subject|json_encode}',
+                'subject'       => 'abcd',
+                'expectedString' => '["abcd"]',
+            ],
+            'withNull' => [
+                'template'      => 'string:{","|explode:$subject|json_encode}',
+                'subject'       => null,
+                'expectedString' => '[""]',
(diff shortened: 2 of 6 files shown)


wisskid:
This seems great at first glance, and all the tests seem to be passing. However, it feels a bit weird to start providing wrappers for native PHP-functions, only to prevent deprecation errors that result from these very PHP-functions. I can see why one might want this, but where is this going to end? Are we going to be writing wrapper plugins to circumvent each and every change in PHP from now on?

mfettig:
For me I simply wasn't able to think of another way to address the reported items in #750. I felt this approach was similar in spirit to changes like #743 in providing a default which preserves existing behavior, but I do see the potential headache that could stem from creating a layer in front of PHP built-ins.

If there is a different approach that would be preferred or if the decision is just not to address these deprecation issues at all then let me know and I'll amend or close the PR as needed.

rudiedirkx:
@wisskid Yes, I think that's what a template engine should do. Whatever version of PHP it runs on, it works the same. However strict PHP decides to get, Smarty decides its own strictness. I think that's only possible with wrappers like this.

wisskid:
@mfettig @rudiedirkx you are both right. I've given it some thought, and I think we should go forward in the direction you propose. One way out of the never ending chore of keeping up with changes in PHP could be to drop support for passthrough of modifiers to native PHP functions in Smarty 5. We could provide a set of standard modifier plugins to cover the most common cases. The ones provided by @mfettig would be a good start, I think.

tests/UnitTests/TemplateSource/TagTests/PluginModifier/PluginModifierExplodeTest.php
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * @return void
+     * @throws \SmartyException
+     *
+     * @dataProvider explodeDataProvider
+     */
+    public function testExplode($template, $subject, $expectedString)
+    {
+        $this->smarty->setDebugging(true);
```
wisskid, line 27: > This enables the [debugging console](https://smarty-php.github.io/smarty/programmers/api-variables/variable-debugging.html#chapter.debugging.console). The console is a javascript popup window that informs you of the [included](https://smarty-php.github.io/smarty/programmers/api-variables/variable-debugging.html#language.function.include) templates, variables [assigned](https://smarty-php.github.io/smarty/programmers/api-variables/variable-debugging.html#api.assign) from php and [config file variables](https://smarty-php.github.io/smarty/programmers/api-variables/variable-debugging.html#language.config.variables) for the current script.

Why enable the debugging console in unit tests?

tests/UnitTests/TemplateSource/TagTests/PluginModifier/PluginModifierNumberFormatTest.php
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * @return void
+     * @throws \SmartyException
+     *
+     * @dataProvider numberFormatDataProvider
+     */
+    public function testNumberFormat($template, $subject, $expectedString)
+    {
+        $this->smarty->setDebugging(true);
```
wisskid, line 27: and here

tests/UnitTests/TemplateSource/TagTests/PluginModifier/PluginModifierExplodeTest.php
mfettig: I don't remember what I was trying to accomplish with that line. I've removed it now

mfettig:
@wisskid I've removed the setDebugging() calls and rebased this branch onto the current master branch to resolve a conflict that had cropped up

tests/UnitTests/TemplateSource/TagTests/PluginModifier/PluginModifierNumberFormatTest.php
```diff
@@ -0,0 +1,59 @@
...
+                'subject'       => 12345,
+                'expectedString' => "12,345",
+            ],
+            'withDecimalDefault'  => [
+                'template'      => 'string:{$subject|number_format}',
+                'subject'       => 12345.6789,
+                'expectedString' => "12,346",
+            ],
+            'withDecimalAndExtras'  => [
+                'template'      => 'string:{$subject|number_format:2:\'-\':\'~\'}',
```
AnrDaemon, line 48: You could use double quotes inside singles to avoid unnecessary quoting.

libs/plugins/modifier.number_format.php
```diff
@@ -0,0 +1,26 @@
...
+ * @param int         $decimals
+ * @param string|null $decimal_separator
+ * @param string|null $thousands_separator
+ *
+ * @return string
+ */
+function smarty_modifier_number_format(?float $num, int $decimals = 0, ?string $decimal_separator = ".", ?string $thousands_separator = ",")
+{
+    // provide $num default to prevent deprecation errors in PHP >=8.1
+    return number_format($num ?? 0.0, $decimals, $decimal_separator, $thousands_separator);
```
AnrDaemon, line 25: Since you are declaring `$decimal_separator` and `$thousands_separator` nullable, you should account for null values as well.

AnrDaemon (changes requested):

tests/UnitTests/TemplateSource/TagTests/PluginModifier/PluginModifierNumberFormatTest.php
mfettig: Updated this to use double quotes

libs/plugins/modifier.number_format.php
mfettig: If that is really wanted I'll add some more null coalesces, but since number_format itself accounts for nulls for those arguments I didn't see any value in doing so

PR #788, opened 2022-08-25
fix upper modifier and strftime usage in date_format if PHP > 8.1

cmelius:
This fixes: 
"|upper" and
"|date_format"    (see #750 )
on PHP8.1

function php8strftime should probably be moved to some function.strftime.php to use with other occurrences of strftime?

### DIFF
--- a/libs/plugins/modifier.date_format.php
+++ b/libs/plugins/modifier.date_format.php
@@ -78,8 +78,198 @@ function smarty_modifier_date_format($string, $format = null, $default_date = ''
             }
             $format = str_replace($_win_from, $_win_to, $format);
         }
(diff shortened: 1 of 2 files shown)


wisskid:
@cmelius yes, the strftime-fix is too specific. `strftime` is also used in `libs/plugins/function.html_select_date.php`. See #672.
The upper changes seem fine. Do you want to update your PR? Or shall I copy your changes for the upper modifier?

cmelius:
You're welcome to copy. I am on vacation for the next two weeks so please go ahead.

Cheers

wisskid:
will do, closing this.

PR #796, opened 2022-09-12
Bugfix/rand without param v3

wisskid:


composer.json
```diff
@@ -27,7 +27,7 @@
         "forum": "http://www.smarty.net/forums/"
     },
     "require": {
-        "php": ">=5.2"
+        "php": "^5.2 || ^7.0"
```
estan, line 30: Hey, I had been using smarty 3.1.x with PHP 8. Just want to confirm that this was intended - not compatible with 8?

I was actually using it indirectly thought the Widgets extension for MediaWiki, which requires smarty 3.1.x. But I ran into a problem trying to install it in a mediawiki:1.39.1 Docker container recently (which uses PHP 8), since the Widgets extension had bumped their smarty requirement to 1.3.47 to get a recent security fix.

(see https://github.com/wikimedia/mediawiki-extensions-Widgets/commit/a13445f4a7e407f9cb440f85020a8bd6fe48fa16#r102066674)

wisskid: Yes, PHP8 support has been added in smarty v4. It should be a relatively cheap upgrade.

marclaporte: And Smarty 5 is in development 👍 
https://github.com/smarty-php/smarty/pull/852

estan: Alright, thanks guys. Now I actually saw that the MediaWiki Widget extensions folks moved back to 3.1.46 in their master branch due to this, I guess until they've ensured it'll work with 4.x (https://github.com/wikimedia/mediawiki-extensions-Widgets/commit/2022ef9bf3f7a771010ac9965b339332a476ca91).

wisskid: I can see why they do this, but in reality the 3.1.46 version is of course no more php8 compatible than 3.1.47 is.

bawolff: Hi, yes. The 3.1.46 thing we did is just super temporary until we can update to 4, which we intend to do in short order once we fix our incompatibilities with 4.x

PR #798, opened 2022-09-14
clean output buffer for Throwable instead of just Exception

wisskid:


wisskid:
I'm not convinced this is safe yet. What happens < PHP7.0 (when Throwable was first introduced). Will the exception handler be triggered?

PR #801, opened 2022-09-16
fix compilation for caching templates

Storyxx:
We maintain a complex template setup using inheritance and includes. 
When activating caching, we get an out of memory error due to a non-terminating recursion. 
After stepping through the code there seem to be two possible causes, which when corrected resolve the issue.

The first one being missing parenthesis around the ternary operator which leads to a wrong comparison.

With the second one it seems that the wrong template was compiled.

PHP-version: 8.0
Smarty-version: 4.2.0


wisskid:
@Storyxx both seem like accidental typos indeed. Wonder how they went unnoticed for so many years.
Anyway, does this fix your problems?

PR #804, opened 2022-09-19
revert php constraints to not force major updates

piranos:
the change in the php constraints forces codebases using php 8 to upgrade to the next major.
this changes should not be made in a patch update

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "forum": "http://www.smarty.net/forums/"
     },
     "require": {
-        "php": "^5.2 || ^7.0"
+        "php": ">=5.2"
     },
     "autoload": {
         "classmap": [


wisskid:
Smarty 3.1 was never meant to aupport php8. Smarty 4 was released for that purpose. This change was meant to clarify this as people would keep requesting fixes for running smarty v3 on php8.

piranos:
i don't think that is a good reason to introduce a compatibility change without even a notice in the changelog.
i think a better place to clarify would be the readme

wisskid:
I agree that it should have been mentioned in the changelog. The readme.md for v:.1 already states "Smarty can be run with PHP 5.2 to PHP 7.4."

wisskid:
The changelog has been updated (in the support/3.1 branch) to clarify.

PR #832, opened 2022-11-24
Allow dereferencing of non-objects accross all supported PHP versions

### LINKED ISSUE #831: Dereferencing of non-object triggers E_WARNING in php >= 8.0
```
$smarty->assign('object', null);
====
{{$object->name}}
```

That would trigger a

> Attempt to read property "name" on null

_Originally posted by @rudiedirkx in https://github.com/smarty-php/smarty/issues/735#issuecomment-1326258423_

wisskid:
Fixes #831


rudiedirkx:
I have no idea what _DereferencingNonObjects_ is, so I would put all those error suppression in the same config, but sure, this works :+1:

rudiedirkx:
What about

```
class Foo {}
$smarty->assign('object', new Foo);
====
{{$object->name}}
```

? Is that caught too? A non-existing property of a valid object? Or a protected/private property of an object, which is probably a different message:

```
class Foo {
  protected $name = 'Bar';
}
$smarty->assign('object', new Foo);
====
{{$object->name}}
```

PR #834, opened 2022-12-15
Fix PHP8.1 deprecation warnings on null string in modifiers

AdrienPoupa:
I noticed that if one is calling `|truncate` on null, it triggers `E_DEPRECATED: mb_strlen(): Passing null to parameter #1 ($string) of type string is deprecated` on PHP 8.1.

This is expected, but given the modifier accepts null we may as well return an empty string before calling the function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `$smarty->muteUndefinedOrNullWarnings()` now also mutes PHP7 notices for undefined array indexes [#736](https://github.com/smarty-php/smarty/issues/736)
 - `$smarty->muteUndefinedOrNullWarnings()` now treats undefined vars and array access of a null or false variables 
   equivalent across all supported PHP versions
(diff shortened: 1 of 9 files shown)


rudiedirkx:
Didn't we do this recently in a huge PR? Or did that not include modifiers & functions? It seems inefficient to do every modifier/function separately like this and https://github.com/smarty-php/smarty/commit/d304d349b4deb19c84742b29b53bb3f8e59413fc and more. On the other hand, I don't even run 8.1 yet, so I'm not going to fix all of them =) so you can safely ignore me.

AdrienPoupa:
I didn't see it on `master` but if it's in a different branch, then for sure this can be closed.

AdrienPoupa:
I added some more changes that were probably left out previously, it should all be good now :)

AdrienPoupa:
We've been using this in production for some now, it works fine. We would love to see that merged and released so we can drop our internal fork :)

wisskid:
No problem! Thank you for your contribution.

AdrienPoupa:
Thanks for the merge!

PR #835, opened 2022-12-15
New use_only_compiled option

gtozzi:
If set to TRUE, Smarty will use only compiled templates, also
ignoring the (in)existence of base templates. Compiled filenames
will be constant and relative to the template basename.
Useful to use pre-compiled templates and distribute them already
compiled on production websites. Overrides and disables
$merge_compiled_includes

### DIFF
--- a/docs/programmers/api-variables.md
+++ b/docs/programmers/api-variables.md
@@ -45,6 +45,7 @@ them directly, or use the corresponding setter/getter methods.
 - [$template_dir](./api-variables/variable-template-dir.md)
 - [$trusted_dir](./api-variables/variable-trusted-dir.md)
 - [$use_include_path](./api-variables/variable-use-include-path.md)
+- [$use_only_compiled](./api-variables/variable-use-only-compiled.md)
 - [$use_sub_dirs](./api-variables/variable-use-sub-dirs.md)
 
 > **Note**
--- a/docs/programmers/api-variables/variable-use-only-compiled.md
+++ b/docs/programmers/api-variables/variable-use-only-compiled.md
@@ -1,0 +1,9 @@
+\$use\_only\_compiled {#variable.use.only.compiled}
+================
+
+If set to TRUE, Smarty will use only compiled templates, alsoù
+ignoring the (in)existence of base templates. Compiled filenames
+will be constant and relative to the template basename.
+Useful to use pre-compiled templates and distribute them already
+compiled on production websites. Overrides and disables
+[`$merge_compiled_includes`](#variable.merge.compiled.includes)
--- a/libs/Smarty.class.php
+++ b/libs/Smarty.class.php
@@ -579,6 +579,13 @@ class Smarty extends Smarty_Internal_TemplateBase
     public $_debug = null;
 
     /**
+     * force using of only compiled templates, ignoring source
+     *
+     * @var boolean
+     */
+    public $use_only_compiled = false;
+
+    /**
      * template directory
      *
      * @var array
@@ -1277,6 +1284,14 @@ class Smarty extends Smarty_Internal_TemplateBase
     public function setCachingType($caching_type)
     {
         $this->caching_type = $caching_type;
+    }
+
+    /**
+     * @param boolean $use_only_compiled
+     */
+    public function setUseOnlyCompiled($use_only_compiled)
+    {
+        $this->use_only_compiled = $use_only_compiled;
     }
 
     /**
--- a/libs/sysplugins/smarty_internal_compile_include.php
+++ b/libs/sysplugins/smarty_internal_compile_include.php
@@ -145,6 +145,10 @@ class Smarty_Internal_Compile_Include extends Smarty_Internal_CompileBase
             if (isset($_attr[ 'compile_id' ]) && $compiler->isVariable($_attr[ 'compile_id' ])) {
                 $merge_compiled_includes = false;
             }
+            // compile names without uid
+            if ($compiler->smarty->use_only_compiled) {
+                $merge_compiled_includes = false;
+            }
         }
         /*
         * if the {include} tag provides individual parameter for caching or compile_id
--- a/libs/sysplugins/smarty_internal_template.php
+++ b/libs/sysplugins/smarty_internal_template.php
@@ -192,7 +192,7 @@ class Smarty_Internal_Template extends Smarty_Internal_TemplateBase
             $this->smarty->_debug->start_template($this, $display);
         }
         // checks if template exists
-        if (!$this->source->exists) {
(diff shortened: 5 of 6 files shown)


gtozzi:
> How does this relate to the compile_check setting?

compile_check requires source template files to be present, use_only_compiled allows the dev to remove them completely: they do not need to be deployed in production.

compile_check compiles the template only once at first run in production, then stops checking about updates. use_only_compiled is meant to compile templates manually in a script before deployment (i.e. in you build CI) and deploy compiled-only templates.

wisskid:
> compile_check requires source template files to be present

You are right. Even with $compile_check enabled, Smarty still checks if source files exist, although it won't compare the timestamps. I stumbled across this earlier this week and I actually think it is either a bug or feature that could be improved.

The way I see it: if you are going to check for the existence of a file, you might as well check its timestamp. Or, if you want to optimize for performance and want don't check the timestamp, then don't check for the existence of the file either.

So the way I see it, with $compile_check enabled, Smarty should **not** check for the existence of the source templates. I wrote the the following unit test to test compile_check behavior:

```php
<?php
/**
 * Smarty PHPunit tests for compile check
 */
class CompileCheckTest extends PHPUnit_Smarty
{
    public $methodName = null;
    public function setUp(): void
    {
        $this->setUpSmarty(__DIR__);
        $this->smarty->addTemplateDir('./templates_tmp');
        $this->cleanDirs();
    }

    /**
     * generate templates
     */
    protected function makeFiles()
    {
        file_put_contents('./templates_tmp/t1.tpl', 'TPL1');
        file_put_contents('./templates_tmp/t2.tpl', 'TPL2');
        file_put_contents('./templates_tmp/base.tpl', '{include file="t1.tpl"}{include file="t2.tpl"}');
    }
    /**
     * remove generated templates
     */
    protected function removeFiles()
    {
        unlink('./templates_tmp/t1.tpl');
        unlink('./templates_tmp/t2.tpl');
        unlink('./templates_tmp/base.tpl');
    }

    /**
     * reset, but leave the files alone
     * @return void
     */
    private function softResetSmarty() {
        $this->smarty = new Smarty();
        $this->smarty->addTemplateDir('./templates_tmp');
    }

    /**
     * @group slow
     */
    public function testCompileCheckOn0()
    {
        $this->makeFiles();
        $this->assertEquals('TPL1TPL2', $this->smarty->fetch('base.tpl'));

        $this->softResetSmarty();
        $this->smarty->setCompileCheck(Smarty::COMPILECHECK_ON);

        $this->assertEquals('TPL1TPL2', $this->smarty->fetch('base.tpl'));
    }

    /**
     * @group slow
     */
    public function testCompileCheckOn1()
    {
        $this->makeFiles();
        $this->assertEquals('TPL1TPL2', $this->smarty->fetch('base.tpl'));

        $this->softResetSmarty();
        $this->smarty->setCompileCheck(Smarty::COMPILECHECK_ON);

        unlink('./templates_tmp/base.tpl');
        sleep(1);

        $this->expectException(Exception::class);
        $this->smarty->fetch('base.tpl');
    }

    /**
     * @group slow
     */
    public function testCompileCheckOn2()
    {
        $this->makeFiles();
        $this->assertEquals('TPL1TPL2', $this->smarty->fetch('base.tpl'));

        $this->softResetSmarty();
        $this->smarty->setCompileCheck(Smarty::COMPILECHECK_ON);

        sleep(1);
        file_put_contents('./templates_tmp/base.tpl', 'hello');

        $this->assertEquals('hello', $this->smarty->fetch('base.tpl'));
    }

    /**
     * @group slow
     */
    public function testCompileCheckOff0()
    {
        $this->makeFiles();
        $this->assertEquals('TPL1TPL2', $this->smarty->fetch('base.tpl'));

        $this->softResetSmarty();
        $this->smarty->setCompileCheck(Smarty::COMPILECHECK_OFF);

        $this->assertEquals('TPL1TPL2', $this->smarty->fetch('base.tpl'));
    }

    /**
     * @group slow
     */
    public function testCompileCheckOff1()
    {
        $this->makeFiles();
        $this->assertEquals('TPL1TPL2', $this->smarty->fetch('base.tpl'));

        $this->softResetSmarty();
        $this->smarty->setCompileCheck(Smarty::COMPILECHECK_OFF);

        unlink('./templates_tmp/base.tpl');
        sleep(1);

        $this->assertEquals('TPL1TPL2', $this->smarty->fetch('base.tpl'));
    }

    /**
     * @group slow
     */
    public function testCompileCheckOff2()
    {
        $this->makeFiles();
        $this->assertEquals('TPL1TPL2', $this->smarty->fetch('base.tpl'));

        $this->softResetSmarty();
        $this->smarty->setCompileCheck(Smarty::COMPILECHECK_OFF);

        sleep(1);
        file_put_contents('./templates_tmp/base.tpl', 'hello');

        $this->assertEquals('TPL1TPL2', $this->smarty->fetch('base.tpl'));
    }

}
```

You'll notice that testCompileCheckOff1 fails. If we fix this, we won't need an extra setting.

Your idea of making the compiled filenames independent of the actual source file names / file path is interesting as well. I am working on that in v5.

gtozzi:
> The way I see it: if you are going to check for the existence of a file, you might as well check its timestamp. Or, if you want to optimize for performance and want don't check the timestamp, then don't check for the existence of the file either.
> 
> So the way I see it, with $compile_check enabled, Smarty should **not** check for the existence of the source templates. I wrote the the following unit test to test compile_check behavior:
> 
> You'll notice that testCompileCheckOff1 fails. If we fix this, we won't need an extra setting.
> 
> Your idea of making the compiled filenames independent of the actual source file names / file path is interesting as well. I am working on that in v5.

We tried this approach in first place but looks like it was too invasive because compiled file names include the hash of the source file.

It looked like a too invasive mod to change the file naming scheme for the whole smarty, so we opted for a less invasive new flag. When enabled, compiled file names no loner include the hash in the name, making possible to have a two-way relation between the source and its compiled result.

wisskid:
Could you try `composer require smarty/smarty:dev-smarty5` and let me know how that works for you?
```php
<?php

use \Smarty\Smarty;
$smarty = new Smarty();
$smarty->setCompileCheck(false);
//...
```
should do the trick

gtozzi:
> Could you try `composer require smarty/smarty:dev-smarty5` and let me know how that works for you?

I'm not using composer, can i checkout the "smarty5" branch?

wisskid:
YesSimonOp 6 feb. 2023 om 12:57 heeft Gabriele Tozzi ***@***.***> het volgende geschreven:﻿

Could you try composer require smarty/smarty:dev-smarty5 and let me know how that works for you?

I'm not using composer, can i checkout the "smarty5" branch?

PR #838, opened 2023-01-03
Fix PHP 8.1 deprecation notice in `truncate` for NULL values

SunMar:
If the input string is a `NULL` value, PHP 8.1 throws the following deprecation notice:
```
mb_strlen(): Passing null to parameter #1 ($string) of type string is deprecated
```
Catching it in the modifier avoids having to add boilerplate `NULL` checks in the template.


SunMar:
Oops, I see #834 was already created for this.

Issue #839, opened 2023-01-05
Error handler should only allow notices etc in smarty templates, not beyond


rudiedirkx:
If the error handler `allow` properties are true (default), Smarty catches ALL notices (in short) for ALL code that STARTS in a Smarty template. That's a bit too much. This notice should be caught like now:

    {{$nonExistingVar}}

That throws an `Undefined array key` from `24555ff3ba201974a27652ea2c61a898c2df0cb8_0.file.bla.tpl` and we want to catch that. That's good.

But this one should NOT be caught by the Smarty error handler:

    {{$existingObject->existingMethod()}}

    function existingMethod() {
        return $this->existingArray['nonExistingKey'];
    }

That also throws an `Undefined array key`, but NOT from compiled template code. But Smarty catches and suppresses it anyway. I don't think that's what we want.

Can `Smarty_Internal_ErrorHandler::handleError()` check for the source of the error? If it's not a smarty compiled template file, let it through.

rudiedirkx:
See #840

PR #840, opened 2023-01-05
Error handler should only allow notices etc in smarty templates, not beyond

rudiedirkx:
Only handle errors in the smarty compiled dir. Not errors that happen during smarty but somewhere else.


rudiedirkx:
See #839

lemon-juice:
This has always been an important missing feature in Smarty error handler. [rudiedirkx](https://github.com/rudiedirkx)'s solution looks good, I hope it gets merged into Smarty!

rudiedirkx:
@wisskid Did you catch this somewhere else by now? I haven't been paying attention to Smarty merges. My method might not be acceptable, but Smarty has to check against `$smarty->getCompileDir()` probably.

wisskid:
@rudiedirkx I hadn't looked into this yet. Seems like a great improvement. At some point, I would like to get rid of the error handler altogether, but this seems like a great first step. I'll have to adapt it to the new v5 release, but this one will do just fine for v4. I'll merge it asap.

rudiedirkx:
The base branch change did something weird to the diff. I resolved a merge conflict, but it's still weird. The error handler has changed in the meantime probably. All I did was add

```php
if (strpos($errfile, $this->smarty->getCompileDir()) === 0) {
```

around all of the `return` conditions in `handleError()`.

----

Smarty still needs its own error handler, doesn't it? You don't want an uncaught error to trace to a compiled Smarty template. Or maybe you do... I don't know. Twig catches all errors during a template and rethrows it with original template info. That's good for debugging.

rudiedirkx:
I can't change the base branch of my fork, so I can't fix this.

wisskid:
@rudiedirkx how about uncompiled templates such as 'eval:' templates or stream plugins?

rudiedirkx:
No idea. I've never used those. What are `$errfile` and `$this->smarty->getCompileDir()` then?

wisskid:
The compile dir is probably still the (default) compile dir, but $errfile will be the phone script calling the template. Which will not be in the compile dir.

Issue #850, opened 2023-01-21
Parent scope in template blocks does not bubble up when it extends another template


maryo:
I'd like to update Smarty in a big legacy project with code like this. It is just a simplfied example which stopped working in `v3.1.28`:

`index.tpl`:
```php
{$test="index"}
{include "./include.tpl"}
test: {$test} {* displays "index", I'd like to propagate "include_base" without polluting the global/root scope *}
```

`include.tpl`:
```php
{extends "./include_base.tpl"}
{*$test=$test scope=parent*} {* this also doesn't help because the variable in `include_base.tpl` is assigned later that this code *}
```

`include_base.tpl`:
```php
{$test="include_base" scope=parent}
```

-----------------------------------

Trying to affect the execution order using blocks

`include.tpl`:
```php
{extends "./include_base.tpl"}

{block content append}
    {$test=$test scope=parent} {* this also doesn't help because the parent of this template is not "index.tpl" *}
    {*$smarty.template_object->parent->parent->_assignInScope("test", $test)*} {* this is the only hack i came up with that works *}
{/block}
```

`include_base.tpl`:
```php
{block content}
    {$test="include_base" scope=parent}
{/block}
```

Is it a bug or an intended behavior? Shouldn't there be a way how to propagate variable to the parent "scope by scope"?.
Our code is legacy and propagation up the tree might even not be a good idea but it is still a BC break that prevents us to update Smarty without fear.

BTW Thanks for everyone who still tries to maintain this.

PR #851, opened 2023-01-23
Bubbling up of parent scope in template blocks with inheritance chain

maryo:
Updating variable in parent scope (when the `scope=parent` is explicitly used on a tag like `assign`) now bubbles up (to the including template) even when the included template is part of an inheritance chain **as it did pre `v3.1.28`**.

Fixes https://github.com/smarty-php/smarty/issues/850

### DIFF
--- a/libs/sysplugins/smarty_internal_runtime_updatescope.php
+++ b/libs/sysplugins/smarty_internal_runtime_updatescope.php
@@ -31,7 +31,7 @@ class Smarty_Internal_Runtime_UpdateScope
                 Smarty::$global_tpl_vars[ $varName ] = $tpl->tpl_vars[ $varName ];
             }
             // update scopes
-            foreach ($this->_getAffectedScopes($tpl, $mergedScope) as $ptr) {
+            foreach ($this->_getAffectedScopes($tpl, $tagScope) as $ptr) {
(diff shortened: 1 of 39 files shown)


libs/sysplugins/smarty_internal_runtime_updatescope.php
```diff
@@ -44,16 +44,29 @@ public function _updateScope(Smarty_Internal_Template $tpl, $varName, $tagScope
...
      */
-    public function _getAffectedScopes(Smarty_Internal_Template $tpl, $mergedScope)
+    public function _getAffectedScopes(Smarty_Internal_Template $tpl, $tagScope)
     {
+        $mergedScope = $tagScope | $tpl->scope;
         $_stack = array();
         $ptr = $tpl->parent;
         if ($mergedScope && isset($ptr) && $ptr->_isTplObj()) {
             $_stack[] = $ptr;
+            if ($tpl->inheritance && $tagScope & Smarty::SCOPE_PARENT) {
```
maryo, line 58: Templates with an inheritance chain have an implicit `Smarty::SCOPE_PARENT`. I don't know the exact reason but the variables should not bubble up outside the inheritance chain implicitly so I have to explicitly test against the tag scope.

Issue #736, opened 2022-02-18
PHP 8 focused error handler should also accept PHP 7 notices


rudiedirkx:
`Smarty_Internal_ErrorHandler::handleError()` only checks for PHP 8 style error strings, but not "Undefined index: abc". I can do that somewhere else, or not listen for notices, but why not here where the same logic for PHP 8 lives.

Probably add inside the `allowUndefinedArrayKeys` block: `Undefined index:`

rudiedirkx:
I'm running the same Smarty 4.1 app on a PHP 8.0 and a PHP 7.4 machine, and the error handler only works as expected on PHP 8.0. I have my app's error handler checking for Smarty notices now =(

Rambomst:
I am also running into this issue, the PR attached resolved the problem. Thanks @rudiedirkx.

Any word on when this will be merged/released @wisskid?

wisskid:
@Rambomst do you have a code example that triggers the error? I'll add a unit test and merge it asap

Rambomst:
> @Rambomst do you have a code example that triggers the error? I'll add a unit test and merge it asap

Honestly, no :(.

We do set the error_reporting value for smarty by doing:
```
$smarty->error_reporting = E_ALL & ~E_NOTICE & ~E_STRICT & ~E_DEPRECATED & ~E_WARNING
```
This is happening in a monolithic application and I was only able to validate this working by manually modifying the vendor file on a dev server, unfortunately I don't have a neat snippet.

I am happy to provide as much information as I can, let me know if you need anything else.

This is happening on PHP 7.2.34 and we are using Sentry error reporting which does some error handling itself.

Sorry I'm not more helpful :(

PR #857, opened 2023-02-02
Fixing the warning of unintialized variables in php 8.1

aamirsofi:
While working on the PHP 8.1 refactoring I came across a common problem/error related to accessing variables in the tpl files that are not set/passed from the controller.
Instead of checking at all places if the variable is set or not, we can fix it by passing the defult NULL value.


gries:
fyi you can also use https://github.com/smarty-php/smarty/issues/736 `$smarty->muteUndefinedOrNullWarnings()`

Issue #633, opened 2021-01-14
nullsafe operator is not working


mikehageman:
Using ?-> results in: Unexpected "?", expected one of: "}" when php 8 is used

wisskid:
@mikehageman could you provide a little more context? E.g. a line of an actual template?

mikehageman:
Sorry ofcourse, I figured it's one of the new key feautures of php8 so it would be kinda obvious (and unfortunately i do not have the time to fix it myself)

> <input id="images_categories"
>        name="images_categories"
>        list="images_categories_data"
>        maxlength="50"
>        class="form-control"
>        value="{$images->get_images_categories()?->get_category()}"
> >
> 
> Fatal error: Uncaught --> Smarty Compiler: Syntax error in template "file:....images.tpl" on line 51 "value="{$images->get_images_categories()?->get_category()}"" - Unexpected "?", expected one of: "}" <-- thrown in ..../vendor/smarty/smarty/libs/sysplugins/smarty_internal_templatecompilerbase.php on line 51

_$images_ is a database model, so if i call _get_images_categories()_ it wil give me either _null_ or a _images_categories_ model. The operator ?-> makes sure that _get_category()_ won't be executed if _get_images_categories()_ evaluates to null. https://stitcher.io/blog/php-8-nullsafe-operator

wisskid:
@mikehageman thank you! We don't suppor the nullsafe operator yet. I'll put it on the backlog, cause it looks very much worth considering.

j-applese3d:
I would be interested in implementing this. 
I assumed I simply needed to add & slightly change the same code that "PTR" uses, which seems to be true.
I don't have any prior experience with yacc fwiw.

I _do_ have a working change, but I'm not sure if I should make a pull request in this repo, or in https://github.com/smarty-php/smarty-lexer ?

Thanks!

Snor:
> I would be interested in implementing this. I assumed I simply needed to add & slightly change the same code that "PTR" uses, which seems to be true. I don't have any prior experience with yacc fwiw.
> 
> I _do_ have a working change, but I'm not sure if I should make a pull request in this repo, or in https://github.com/smarty-php/smarty-lexer ?

I've been looking at this too as I've wanted this feature in Smarty for years - long before it was even in PHP! It's insanely useful and reduces template code/increases readability massively!

[...]

PR #860, opened 2023-02-08
Add support for PHP 8s nullsafe operator. #633

j-applese3d:
I'm not sure that I got everything right, so please review and let me know what I need to change :)

For instance the test will fail if you run it with PHP <8.0, and I wasn't certain that specific test file was the right one to update or if I should make a new one. I didn't know where to put it...

Thanks!


wisskid:
@j-applese3d is there any way to compile this into PHP7-compatible PHP code? I.e. to implement the nullsafe operator in Smarty without actually using the nullsafe operator in the compiled code? If not, we'll have to drop support for PHP7.x and we cannot release this until v6.

j-applese3d:
I believe that there are 2 different ways `?->` can be used:
1. to read an object property
2. to call a method

Both of these cases should be easy enough to implement in PHP compatible with 7.0.
Here are some example functions:

```php
/** {$obj?->doSomething($p, $p2)} ➡ nullSafeMethod($obj, "doSomething", [$p, $p2]); */
function nullSafeMethod($object, $function, ...$params)
{
    if ($object === null) {
        return null;
    }
    return $object->{$function}($params);
}

/** {$obj?->someProperty} ➡ nullSafeProperty($obj, "someProperty"); */
function nullSafeProperty($object, $property)
{
    if ($object === null) {
        return null;
    }
    return $object->{$property};
}
```

wisskid:
I'm trying to implement your suggestion, but it requires significant changes in the parser. While at it, I'm wondering if it would be even better to make `->` nullsafe by default in Smarty. And also array member access.  That way, we could ditch the custom error handler used to suppress warnings for "undefined index", "undefined array key" and "trying to read property of null".

j-applese3d:
> it requires significant changes in the parser.

Yes... I did try to fix my solution, but I couldn't find where/how to insert the custom logic (as opposed to simply mapping `?->` to `?->`)


> I'm wondering if it would be even better to make -> nullsafe by default in Smarty

I guess it depends on the main purpose of Smarty.
In my situation, I code the backend and frontend, so I think of Smarty as being syntactic sugar, and a nice way to separate my "View".

Coming from this point of view, it _doesn't_ make sense to magically treat `->` as if it were `?->`, rather the developer (templater :slightly_smiling_face:) should receive the same error as they would in PHP. This lets them know they made a mistake, and if they want `?->` they can adjust it; but maybe they forgot to assign the variable -- or used the wrong name.

> Call to a member function [function name]\() on null

However, I understand that I'm not the only person using this library. [1]

----

**TL;DR:** I like the idea of ditching the custom error handler, but I'd prefer if the "undefined index" errors show up the same as when PHP runs into them. (i.e. leave error handling up to PHP)

----

[1] I would be curious to know how many projects have a team of template editors that _cannot_ edit the backend code, as it feel like many things are directed toward preventing unauthorized execution php code. For instance, deprecating usage of `implode`: for a php dev it is the name they are familiar with. Using `|join:', '` makes sense, it's just not very natural to use `join` in templates when I use `implode` in php.

Thank you for all the work you put into this project. It is much appreciated. :heart:

Issue #813, opened 2022-09-22
Drop support for using PHP functions as modifiers


wisskid:
In order to be able to provide a stable interface, we should drop support for using PHP functions (actually, any callable) as a modifier and require that any modifier is either part of Smarty by default or added as a plugin. If not, we will continue to have to chase every change PHP decides to make and wrap it in backwards compatibility layers.

wisskid:
#814 was incorrectly marked to close this. It's only the first step.

timmit-nl:
Hi,

We see the deprecated notices and we see we use a small list not jet covered:
                'array_reverse',
                'array_slice',
                'empty',
                'floatval',
                'htmlentities',
                'is_array',
                'json_encode',
                'md5',
                'print_r',
                'strpos',
                'strstr',
                'substr',
                'ucfirst',
                
Some need to be changed in the templates to use for example {md5($var)} instead of {$var|md5} but some we just need or make sense as modifier. 

What is the best practice to make it ("modifier" vs "modifiercompiler")? Then we can make some and push those to smarty.

wisskid:
You can just register them as custom modifier plugins, no need to push them to Smarty.

timmit-nl:
I understand we can make them as custom privately only for us (we already do that for other modifiers), but sharing is caring.

wisskid:
That is much appreciated, but part of the reason we are doing this is to reduce the responsibility of Smarty to maintain backward compatibility on all of PHPs functions. If we add these wrappers, this responsibility is reintroduced too.

timmit-nl:
I understand your standpoint, but some of these modifier are normal to use in templates. Thinking of:
'strpos',
'strstr',
'substr',
'ucfirst',
'htmlentities',

So completly remove those as default is cripling smarty in my opinion.

lkppo:
@timmit-nl 

I never used 'substr' and 'ucfirst'. i don't see them as important.

And the correct thing should be 'mb_substr' in some case, so it's a bad idea to reintroduce them.

Registering is ok.

scottchiefbaker:
I agree with @wisskid that these are better served as separate modifier plug-ins. Perhaps a `contrib` folder could be added to the repo for third party (non-supported) add-ons?

xtcodecom:
> I agree with @wisskid that these are better served as separate modifier plug-ins. Perhaps a `contrib` folder could be added to the repo for third party (non-supported) add-ons?

+1
[...]

PR #871, opened 2023-04-03
Remove `md5` modifier from debug.tpl

j-applese3d:
Replaced with a regular function call. 
See https://github.com/smarty-php/smarty/issues/813

### DIFF
--- a/libs/debug.tpl
+++ b/libs/debug.tpl
@@ -167,9 +167,7 @@
     </html>
 {/capture}
 <script type="text/javascript">
-    {$id = '__Smarty__'}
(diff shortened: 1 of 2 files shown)


wisskid:
This changes the syntax, but not the dependency. It will still fail in future Smarty versions.

j-applese3d:
I have the same question as LDerkzenSawiday. I thought https://github.com/smarty-php/smarty/issues/813 says that `|<phpfunction>` was being removed, not the ability to use PHP functions altogether...

Nevertheless, I have changed it so `md5` is now called by `Smarty_Internal_Debug->display_debug` rather than within the template.

PR #880, opened 2023-04-25
Using PHP functions and static class methods in expressions now also triggers a deprecation notice

### LINKED ISSUE #813: Drop support for using PHP functions as modifiers

wisskid:
... because we will drop support for this in the next major release

Fixes #813

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed
-- Using PHP functions in expressions now also triggers a deprecation notice because we will drop support for this in the next major release [#813](https://github.com/smarty-php/smarty/issues/813)
+- Using PHP functions and unregistered static class methods in expressions now also triggers a deprecation notice because we will drop support for this in the next major release [#813](https://github.com/smarty-php/smarty/issues/813)
 
 ## [4.3.1] - 2023-03-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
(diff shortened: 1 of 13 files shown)


vojtasvoboda:
Just a quick question, the following code will be supported in the next version of Smarty or not?

```
{foreach item=column from=range('A', 'Z')}
    ...
{/foreach}
```

Works with Smarty 4.3.4.

Thanks

wisskid:
> Just a quick question, the following code will be supported in the next version of Smarty or not?
> 
> ```
> {foreach item=column from=range('A', 'Z')}
>     ...
> {/foreach}
> ```

Yes, Smarty 5 still supports the 'old' foreach syntax, @vojtasvoboda .

vojtasvoboda:
The question was about calling the native PHP function range() (and also about other native PHP functions), sorry for the incomplete description :-) @wisskid

wisskid:
> The question was about calling the native PHP function range() (and also about other native PHP functions), sorry for the incomplete description :-) @wisskid
> 
> ```
> {foreach range('A', 'Z'} as $column}
>     ...
> {/foreach}
> ```

Well, my answer wasn't very complete either... :-)
The range function is currently not implemented in v5.0.0-rc2 so it would have to be added as a custom extension. But if you feel that it is important, I would be open to add it to the DefaultExtension set. @vojtasvoboda

vojtasvoboda:
But the range() function is not implemented in v4.3.4 either and works well (i didn't find any range mention in the whole repository) :-) I thought Smarty just call global functions (and also native PHP functions) without any extra definition.

For example when I create a custom global function my_custom_strtoupper(string $name) and call:

```
{my_custom_strtoupper('Vojta')}
```

It works.

So in Smarty 5.x.x calling existing functions (and also native PHP functions) will be disabled?

wisskid:
Yes, that is correct

KarelWintersky:
how to disabled this?

magrigry:
Can you explain what is the correct way for accessing constants ?

magrigry:
Could you tell us what is the recommended way for accessing constants or class constants ?

PR #892, opened 2023-08-04
2.6 - Fix e-notice in modifier.escape.php

colemanw:
A `null` value will cause e-notices in most of the esc_types. Functions like `htmlspecialchars`, `str_replace`, `preg_replace` etc do not strictly allow `null` as an argument.
Using an early-return if the value is `empty()` should also give a subtle performance boost, as none of the esc_types do anything meaningful to empty strings or the number 0.

### DIFF
--- a/libs/plugins/modifier.escape.php
+++ b/libs/plugins/modifier.escape.php
@@ -21,6 +21,11 @@
  */
 function smarty_modifier_escape($string, $esc_type = 'html', $char_set = 'ISO-8859-1')
 {
+    // Early return for null, '', or '0', none of which need to be escaped and null will cause e-notices.
+    if (empty($string)) {
+        return (string) $string;
+    }
+
     switch ($esc_type) {
         case 'html':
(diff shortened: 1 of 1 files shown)


scottchiefbaker:
Could this be simplified to `return "";`?

colemanw:
> Could this be simplified to `return "";`?

No because `'0'` is also `empty()`.

scottchiefbaker:
Ah good point.

colemanw:
As far as I can see there is no benefit to escaping integers either, and I'm not sure how the various escape functions that expect a string will cope with an integer (if not currently, maybe in the future they would thrown an e-notice?) so IMO this could be further improved as
```
if (empty($string) || is_int($string)) {
    return (string) $string;
}
```
But I didn't do that because this is a legacy project so I went for the minimal patch.

wisskid:
Smarty v2 really isn't supported anymore. Don't run it in production, it probably has a bunch of security issues.

PR #902, opened 2023-09-12
date_format: check if already exists

kornrunner:
Since everything in `libs/` gets autoloaded by composer - there's no easy way to override built-in modifiers.

This is an attempt at a workaround for `date_format` specifically, as seen within [smarty_mb_str_replace](https://github.com/smarty-php/smarty/commit/8842e79107dd588fba782ca335133a3f0f70de98#diff-a07f3c868273f233caa2af211448d7b1bd52dd3eb7505f6f98543840d4108517R8)

### DIFF
--- a/libs/plugins/modifier.date_format.php
+++ b/libs/plugins/modifier.date_format.php
@@ -5,82 +5,84 @@
  * @package    Smarty
  * @subpackage PluginsModifier
  */
-/**
- * Smarty date_format modifier plugin
- * Type:     modifier
- * Name:     date_format
- * Purpose:  format datestamps via strftime
- * Input:
- *          - string: input date string
- *          - format: strftime format for output
- *          - default_date: default date if $string is empty
- *
- * @link   https://www.smarty.net/manual/en/language.modifier.date.format.php date_format (Smarty online manual)
- * @author Monte Ohrt <monte at ohrt dot com>
- *
- * @param string $string       input date string
- * @param string $format       strftime format for output
- * @param string $default_date default date if $string is empty
- * @param string $formatter    either 'strftime' or 'auto'
- *
- * @return string |void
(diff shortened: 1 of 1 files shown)


rudiedirkx:
> Since everything in `libs/` gets autoloaded by composer

Is that true? I don't think that's true. My project has Smarty modifiers and functions overridden. Smarty loads those files on demand, not always by Composer. Just add a `plugins_dir` before Smarty's and that's where Smarty will look first.

		$smarty->plugins_dir = [
			PROJECT_INC_SMARTY . '/plugins', // Mine
			PROJECT_VENDOR . '/smarty/smarty/libs/plugins', // Smarty's built-in
		];

kornrunner:
> > Since everything in `libs/` gets autoloaded by composer
> 
> Is that true? I don't think that's true. My project has Smarty modifiers and functions overridden. Smarty loads those files on demand, not always by Composer. Just add a `plugins_dir` before Smarty's and that's where Smarty will look first.

Specified within:  https://github.com/smarty-php/smarty/blob/support/4.3/composer.json#L35-L38

> ```
> 	$smarty->plugins_dir = [
> 		PROJECT_INC_SMARTY . '/plugins', // Mine
> 		PROJECT_VENDOR . '/smarty/smarty/libs/plugins', // Smarty's built-in
> 	];
> ```

That's what I was using, but I get compile errors for using already declared function (because autoloader loaded the Smarty file, even before use).

rudiedirkx:
> Specified within: https://github.com/smarty-php/smarty/blob/support/4.3/composer.json#L35-L38

That's a class map, not functions. Functions aren't autoloaded. Maybe something else includes that file twice. Or a smarty/plugin cache. Or something. I've never had a problem with this. Buuut I have no idea which version I'm running. It would be very weird if this had changed on purpose.

PR #910, opened 2023-10-01
Fix case-sensitive tag names (#907)

### LINKED ISSUE #907: Smarty 5: `registerPlugin()` no longer working with names containing non-lowercase characters
This should output "Test" but gives the following error in Smarty 5 / dev-master. Working with latest 4.x version.

> Smarty\CompilerException: Syntax error in template "string:{customTag}" on line 1 "{customTag}" unknown tag 'customtag' in string:{customTag} on line 1

```php
<?php

require 'vendor/autoload.php';

use Smarty\Smarty;
//use \Smarty;

class Functions
{
    public static function customTag()
    {
        return 'Test';
    }
}

$smarty = new Smarty();
$smarty->registerPlugin(Smarty::PLUGIN_FUNCTION, 'customTag', [Functions::class, 'customTag']);
$smarty->display('string:{customTag}');
```

Jack-Dane:
This should fix #907. 

All tests pass, I thought about adding more tests for various plugins but it seemed overkill, happy to add them if the reviewer thinks otherwise. 

There might be a regression here (I don't think so) but just so everyone is aware this will start failing: 
```
$smarty = new Smarty();
$smarty->registerPlugin(Smarty::PLUGIN_FUNCTION, 'customtag', [Functions::class, 'customTag']);
$smarty->display('string:{customTag}');
```

I have kept the "built-in" tags case-insensitive because they have been relying on them being lowercase, so I think it is sensible to keep it this way.

### DIFF
--- a/src/Compiler/Template.php
+++ b/src/Compiler/Template.php
@@ -464,7 +464,7 @@ class Template extends BaseCompiler {
 	public function compileTag($tag, $args, $parameter = []) {
 		$this->prefixCodeStack[] = $this->prefix_code;
 		$this->prefix_code = [];
-		$result = $this->compileTag2(strtolower($tag), $args, $parameter);
+		$result = $this->compileTag2($tag, $args, $parameter);
 		$this->prefix_code = array_merge($this->prefix_code, array_pop($this->prefixCodeStack));
 		return $result;
 	}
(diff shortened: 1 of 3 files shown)


wisskid:
Seems perfect @Jack-Dane ! The possible regression you are mentioning is not a regression: it also fails in Smarty v4.

PR #930, opened 2024-01-26
Fix incorrect code when escape_html=true

### LINKED ISSUE #928: escape_html generates incorrect php code
When setting `escape_html=true`, the following code:

```
{$foo+$bar}
```

becomes:

```
<?php echo htmlspecialchars((string) $_smarty_tpl->tpl_vars['foo']->value+$_smarty_tpl->tpl_vars['bar']->value, ENT_QUOTES, 'UTF-8');?>
```

The code: `(string) $a + $b` is not correct, maybe it should be `(string) ($a + $b)`.

wxiaoguang:
Fix #928

### DIFF
--- a/src/Compile/PrintExpressionCompiler.php
+++ b/src/Compile/PrintExpressionCompiler.php
@@ -84,7 +84,7 @@ class PrintExpressionCompiler extends Base {
 				}
 
 				if ($compiler->getTemplate()->getSmarty()->escape_html) {
-					$output = "htmlspecialchars((string) {$output}, ENT_QUOTES, '" . addslashes(\Smarty\Smarty::$_CHARSET) . "')";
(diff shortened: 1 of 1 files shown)


wxiaoguang:
Hi @wisskid, is there any interest in this fix? 4.x also needs this fix IMO.

wisskid:
I think so. Can you clarify the actual error this causes?

wxiaoguang:
The function: `htmlspecialchars(string $string, ....)`, the first argument's type is string (PHP is more and more strict about typing)

But `{$foo+$bar}` would result in non-string type:

```
$ php -r 'echo gettype( (string)1+2 );'
integer%
```

So, it needs to use `htmlspecialchars((string) ({$output}), ` to make sure the type is string:

```
$ php -r 'echo gettype( (string)(1+2) );'
string%
```

PR #933, opened 2024-02-14
Deprecate stream variables

wisskid:


marclaporte:
Related: https://github.com/smarty-php/smarty/discussions/884

PR #985, opened 2024-04-05
Throw deprecation notice about unregistered functions without changing the compilation flow

### LINKED ISSUE #964: Version 4.5.1 breaking change: Variables by reference
This is similar to https://github.com/smarty-php/smarty/issues/961, but I couldn't tell if that's specific to version 5 or not.

Smarty 4.5.1 introduces a breaking change regarding parameter references. For example, the following code in Smarty 4.4.1 works:
```smarty
{assign var="match" value=null}
{if preg_match('/([a-z]{4})/', 'a test', $match)}
	{$match.1}
{/if}
```
This will output "test".

However in 4.5.1, it outputs this:
```smarty
Warning: preg_match(): Argument #3 ($matches) must be passed by reference, value given in ...
```

This seems related to the change in `libs/sysplugins/smarty_internal_templatecompilerbase.php` from https://github.com/smarty-php/smarty/commit/c7a271323b9ae4c876c600d360c8555131174264.

wisskid:
Fixes #964

### DIFF
--- a/libs/sysplugins/smarty_internal_templatecompilerbase.php
+++ b/libs/sysplugins/smarty_internal_templatecompilerbase.php
@@ -640,17 +640,18 @@ abstract class Smarty_Internal_TemplateCompilerBase
                         return $func_name . '(' . $parameter[ 0 ] . ')';
                     }
                 } else {
-					$first_param = array_shift($parameter);
-					$modifier = array_merge(array($name), $parameter);
-					// Now, compile the function call as a modifier
-					return $this->compileTag(
-						'private_modifier',
-						array(),
-						array(
-							'modifierlist' => array($modifier),
-							'value'        => $first_param
-						)
-					);
+
+					if (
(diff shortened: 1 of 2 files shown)


libs/sysplugins/smarty_internal_templatecompilerbase.php
```diff
@@ -640,17 +640,18 @@ public function compilePHPFunctionCall($name, $parameter)
...
-						array(),
-						array(
-							'modifierlist' => array($modifier),
-							'value'        => $first_param
-						)
-					);
+
+					if (
+						!$this->smarty->loadPlugin('smarty_modifiercompiler_' . $name)
+						&& !isset($this->smarty->registered_plugins[Smarty::PLUGIN_MODIFIER][$name])
```
wxiaoguang, line 646: Why it only checks `PLUGIN_MODIFIER` but not `PLUGIN_FUNCTION`?

Do I need to register `sprintf` (for example) as modifier and function at the same time?

PR #992, opened 2024-04-11
Fix docblock types

kynx:
This PR fixes a couple of `UndefinedDocblockClass` issues that Psalm was throwing after the upgrade to version 5.

@wisskid Nice work on the new version 😄 I know you've caught some flack for some of the changes, but they are all for the good. We payed attention to the deprecation notices in v4 so it's gone smoothly (so far 🤞)


wisskid:
Thanks!

Issue #994, opened 2024-04-13
getTemplateVars - why the change in return signature in smarty 5?


demeritcowboy:
Short version: Any code that calls getTemplateVars() breaks with smarty5 because it's expecting an array of values, and instead getting an array of objects.

Slightly longer version: At the very least the change seems inconsistent. If you pass in a variable name, it returns its value not the object. Either it should always return an object, or not be a breaking change and do what it did before.

What was the reason for the change, and should it be made more consistent (either way)?

wisskid:
You are correct, this was never intended. I've created https://github.com/smarty-php/smarty/pull/995
Does that fix this issue?

demeritcowboy:
Yes it does. Great! Thank you.

PR #995, opened 2024-04-13
Fix change in signature of getTemplateVars

wisskid:
See #994


eileenmcnaughton:
It would be good to get this tagged in a release - we are seeing further cases of people hitting this so would be nice to get the fix out there

wisskid:
@eileenmcnaughton I've released this in v5.1.0

eileenmcnaughton:
great - thanks @wisskid

Issue #1011, opened 2024-05-06
Smarty v5 built-in json_encode modifier doesn't take Smarty $_CHARSET encoding into account, and can't be overridden using registerPlugin()


cmanley:
I manage a couple of websites, some of which use cp1252 encoding. 
It's in these websites, the json_encode modifier sometimes returns false when given cp1252 encoded data.
This used to work with Smarty v3 (and I think v4 too) where I registered my own json_encode modifier that does the input encoding conversion automatically based on the Smarty $_CHARSET encoding. 
After some debugging, I noticed that the the custom and registered modifier isn't being called at all in v5, but instead the built-in Smarty json_encode() modifier is being used.

I've tried to unregister the built-in json_encode modifier first before registering my custom modifier but that doesn't seem to have any effect:
```
$smarty->unregisterPlugin('modifier',  'json_encode');
$smarty->unregisterPlugin('function', 'json_encode');
$smarty->registerPlugin('modifier', 'json_encode', [$this, 'modifier_json_encode']);
```

If the json_encode modifier is to ignore the $_CHARSET encoding by design, then how do I override a built-in modifier/function with a customized one in v5?

wisskid:
You can find clues on how to do this in #1006

wisskid:
TL;DR. Create a custom Extension.  Register it before the BCPluginAdapter Extension.

wisskid:
But if you consider this to be a bug in the default implementation, we should probably fix that instead?

cmanley:
Thanks. I'll give the setExtension() route a try. I wish it were somewhat easier though.
B.t.w. I consider it a bug, but it also depends on what you consider it to be.

cmanley:
$smarty->setExtensions([
			new \Smarty\Extension\CoreExtension(),
			new \Smarty\Extension\DefaultExtension(),
		]);

wisskid:
I'd be happy to have a look, but I need some more specifics. Can you share a reproduction scenario and your fixed version of the modifier?

cmanley:
Wow, I got it to work using 2 anonymous classes.
$this is in a child class of \Smarty\Smarty here:
```
parent::__construct();
$this->setExtensions([
			new \Smarty\Extension\CoreExtension(),
			new class extends \Smarty\Extension\Base {
				public function getModifierCompiler(string $modifier): ?\Smarty\Compile\Modifier\ModifierCompilerInterface {
					if ($modifier == 'json_encode') {
[...]

PR #1016, opened 2024-05-14
Fix for issue 1011: add 3rd input encoding parameter to json_encode modifier let it default to \Smarty\Smarty::$_CHARSET

cmanley:
See https://github.com/smarty-php/smarty/issues/1011


src/Extension/DefaultExtension.php
```diff
@@ -605,6 +605,98 @@ public function smarty_modifier_join($values, $separator = '')
...
+		}
+
+		# json_encode() expects UTF-8 input, so recursively encode $value if necessary into UTF-8
+		if ($value && strcasecmp($input_encoding, 'UTF-8')) {
+			if (is_string($value)) {	# shortcut for the most common case
+				$value = mb_convert_encoding($value, 'UTF-8', $input_encoding);
+			}
+			elseif (is_array($value) || is_object($value)) {
+				static $transcoder;	# this closure will be assigned once, and then persist in memory
+				if (is_null($transcoder)) {
```
wisskid, line 633: You probably did this because of [this comment](https://github.com/smarty-php/smarty/issues/1011#issuecomment-2099373912) but I'm afraid this is a bit much. Can you refactor the transcoder into a class file under `src/Extension/DefaultExtension/DeepTranscoder.php` or something like that?

src/Extension/DefaultExtension.php
```diff
@@ -605,6 +605,98 @@ public function smarty_modifier_join($values, $separator = '')
...
+						elseif (is_string($data)) {
+							return mb_convert_encoding($data, $to_encoding, $from_encoding);
+						}
+						elseif (is_scalar($data)) {
+							return $data;
+						}
+
+						# convert object to array if necessary
+						if (is_object($data)) {
+							if (is_a($data, '\JsonSerializable')) {	# this is the only reason why this function is not generic
```
wisskid, line 656:
```suggestion
							if (is_a($data, \JsonSerializable::class)) {	# this is the only reason why this function is not generic
```

src/Extension/DefaultExtension.php
```diff
@@ -605,6 +605,98 @@ public function smarty_modifier_join($values, $separator = '')
...
+				} # / if is_null($transcoder)
+
+				$value = $transcoder($value, 'UTF-8', $input_encoding);
+				if ($value === false) {
+					return $value;	# failure; this must not be passed to json_encode!; this is part of what the !empty() check is for at the top of this block
+				}
+			} # / elseif (is_array($value) || is_object($value))
+		} # / if input encoding != UTF-8
+
+		return \json_encode($value, $flags);	# string|false
```
wisskid, line 697: My IDE tells me that \json_encode (and \JsonSerializable::class) are not part of the core of PHP prior to PHP8. We should probably check for `function_exists('json_encode')` and do something accordingly.

cmanley: It's strange that someone would want to use the |json_encode modifier without having the json extension installed. But what for exception (and message) to you suggest to throw if it's not present? The check can be done once (using a static boolean var) the first time the modifier is called.

src/Extension/DefaultExtension.php
cmanley: Looks good too, especially if that helps IDEs check code.

src/Extension/DefaultExtension.php
cmanley: Alright, perhaps DeepJsonTranscode.php then since it's specific to transcoding for json_encode() as indicated by a comment in the code where JsonSerializable is checked.

cmanley: I just pushed a new commit where the json_encode() modifier uses the new class src/Extension/DefaultExtension/RecursiveTranscoder.php, and added a little to the json_encode unit tests.

PR #1030, opened 2024-06-09
Improvement of auto-escaping

Amaury:
This evolution improves the auto-escaping feature.

- The `escape` modifier has no effect when auto-escaping is enabled (when no escape format is given, or when the `html` format is used), to prevent double-escaping.
- The other formats of the `escape` modifier (`htmlall`, `url`, `urlpathinfo`, `quotes`, `javascript`) are processed as we may expect, without double-escaping.
- The new `force` format of the `escape` modifier allows to force double-escaping if needed.
- The new `raw` modifier temporary disables auto-escaping for the expression it is used on.

This Pull Request contains the source code of this evolution, as well as its documentation and unit tests.

### DIFF
--- a/docs/designers/language-modifiers/language-modifier-raw.md
+++ b/docs/designers/language-modifiers/language-modifier-raw.md
@@ -1,0 +1,8 @@
+# raw
+
+Prevents variable escaping when [auto-escaping](../../api/configuring.md#enabling-auto-escaping) is activated.
+
+## Basic usage
+```smarty
+{$myVar|raw}
+```
--- a/src/Smarty.php
+++ b/src/Smarty.php
@@ -416,6 +416,13 @@ class Smarty extends \Smarty\TemplateBase {
 	public $escape_html = false;
 
 	/**
+	 * disabled autoescape (when set to true, the next variable output is not escaped)
+	 *
+	 * @var boolean
+	 */
+	public $raw_output = false;
+
+	/**
 	 * start time for execution time calculation
 	 *
 	 * @var int
--- a/src/Smarty.php
+++ b/src/Smarty.php
@@ -416,13 +416,6 @@ class Smarty extends \Smarty\TemplateBase {
 	public $escape_html = false;
 
 	/**
-	 * disabled autoescape (when set to true, the next variable output is not escaped)
-	 *
-	 * @var boolean
-	 */
-	public $raw_output = false;
-
-	/**
 	 * start time for execution time calculation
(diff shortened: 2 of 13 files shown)


wisskid:
This is looking great, @Amaury !

How do you propose we release this? Do you think this calls for a new major version, i.e. Smarty v6 or a mintor (v5.4)?

One might argue this requires a major, because behavior is changing for existing templates that use `|escape` of some form when html auto-escaping is also enabled. On the other hand: that would very likely be unintended anyway.

docs/designers/language-modifiers/language-modifier-raw.md
```diff
@@ -0,0 +1,8 @@
+# raw
```
wisskid, line 1: This file needs to be added to `/mkdocs.yml` just above 'regex_replace'.

Amaury: OK, I'll do that quickly.

About the release, I would say that a minor version should be enough; I don't know if anybody if really using auto-escaping and `|escape` at the same time. On the other hand, it's true this is a deep change of behaviour…
I don't know, I let you choose ^^'

Amaury:
About the release, I would say that a minor version should be enough; I don't know if anybody if really using auto-escaping and `|escape` at the same time. On the other hand, it's true this is a deep change of behaviour…
I don't know, I let you choose ^^'

src/Smarty.php
```diff
@@ -415,6 +415,13 @@ class Smarty extends \Smarty\TemplateBase {
 	 */
 	public $escape_html = false;
 
+	/**
+	 * disabled autoescape (when set to true, the next variable output is not escaped)
+	 *
+	 * @var boolean
+	 */
+	public $raw_output = false;
```
wisskid, line 423: Just one more thing: I believe we should move this to `Smarty\Compiler\Template`. Placing it in Smarty\Smarty as a public property exposes it to the users and suggests this is a setting one case use instead of a compile time flag. If you want I'll do it for you.

Amaury: I understand and I agree.
And yes, I think you will do it better than me ^^'

wisskid: Cool, did it.

Amaury:
Thank you! :)

Amaury:
In the end, do you think you will create a major version or a minor one?

wisskid:
I'm thinking minor.

PR #1040, opened 2024-07-05
Avoids deprecation notice when using class constants

### LINKED ISSUE #1028: Smarty registered classes check prevents use of class constants to avoid typo bugs in templates
My team makes extensive use of PHP class constants in our Smarty templates to avoid problems with typos in logic checks.  With the new requirement that all classes be registered to access them statically our templates now generate deprecation warnings for each class constant.

It would be nice if class constants references (as opposed to static method calls) did not require class registration.

Alternatively I would like a supported way of overriding this behavior in a security policy.  For example I would expect that overriding `isTrustedStaticClass()` or `isTrustedStaticClassAccess()` in my security policy would allow me to suppress the registered class requirement.  However since the check for class registration is done outside the security policy this does not work unless the security policy also registers the class before returning.  Calling `Smarty::registerClass()` from inside my security policy currently works but does not seem like it is a supported solution to the problem.

wisskid:
Fixes #1028


mkurzydlowski:
@wisskid, will you be releasing this in the next version 4 release?

humpfhumpf:
This does be merged in 4.x, no?

Issue #1007, opened 2024-05-01
CPU spikes in prod every 5 minutes with Smarty v5


francislavoie:
Last week, we tried deploying an update to our prod servers upgrading from Smarty v4.3 to v5.1. Soon after, we started seeing patterns in our CPU graphs showing 100% CPU spikes (for about 5-10 seconds at a time) about every 5 minutes.

We then reverted the change, confirmed that the Smarty version change was the problem. Then, we tried upgrading to v4.4.1 then v4.5.2, to see if those exhibited the behaviour, but they didn't.

Our setup for Smarty is very, very simple. We're not using much in terms of config. It just looks like this (we're running on PHP 8.1 currently):

```php
        // Initialize Smarty
        $smarty = new Smarty();
        $smarty->setTemplateDir("{$basedir}/smarty/templates");
        $smarty->setCompileDir("{$basedir}/smarty/templates_c");
        $smarty->setCaching(Smarty::CACHING_OFF);

        // Plugins, see https://github.com/smarty-php/smarty/issues/813
        $smarty->registerPlugin('modifier', 'json_encode', 'json_encode');
        $smarty->registerPlugin('modifier', 'htmlentities', 'htmlentities');
        $smarty->registerPlugin('modifier', 'htmlspecialchars', 'htmlspecialchars');
        $smarty->registerPlugin('modifier', 'ucfirst', 'ucfirst');
        $smarty->registerPlugin('modifier', 'count', 'count');
        $smarty->registerPlugin('modifier', 'is_object', 'is_object');
        $smarty->registerPlugin('modifier', 'month_names', 'month_names');
        $smarty->registerPlugin('modifier', 'stripos', 'stripos');
        $smarty->registerPlugin('modifier', 'strpos', 'strpos');
        $smarty->registerPlugin('modifier', 'is_null', 'is_null');
        $smarty->registerPlugin('modifier', 'min', 'min');
        $smarty->registerPlugin('modifier', 'max', 'max');
        $smarty->registerPlugin('modifier', 'file_exists', 'file_exists');

        $smarty->registerPlugin('function', 'helper', self::helper(...));
        $smarty->registerPlugin('function', 'popup', self::popup(...));
        $smarty->registerPlugin('function', 'popup_init', self::popup_init(...));
        $smarty->registerPlugin('function', 'querystring', self::querystring(...));
        $smarty->registerPlugin('function', 'selector', self::selector(...));
[...]

PR #1047, opened 2024-07-27
Fixing forced OpCache invalidation on every template include, which is resulting in fast raising wasted OpCache memory #1007

stephanlueckl:
Fixes: https://github.com/smarty-php/smarty/issues/1007

\Smarty\Template\Compiled::loadCompiledTemplate invalidates the cache entry of a compiled template on every call (or include) as it is forced to do so.
This results in fast growing wasted memory and after reaching the tipping point in a total cleanup and refresh of the opcache.

I do not think, that is necessary to force it, seeing this change as hotfix.
 
Nice article about wasted memory of the opcache and it's refreshment mechanism: 
https://www.codana.be/en/insights/php-opcache-realpath-cache-and-preloading

### DIFF
--- a/src/Resource/FilePlugin.php
+++ b/src/Resource/FilePlugin.php
@@ -56,12 +56,14 @@ class FilePlugin extends BasePlugin {
 	 * @param Source $source source object
 	 */
 	public function populateTimestamp(Source $source) {
-		$path = false;
-		if (!$source->exists && $path = $this->getFilePath($source->name, $source->getSmarty(), $source->isConfig)) {
-			$source->timestamp = $source->exists = is_file($path);
+		$path = $this->getFilePath($source->name, $source->getSmarty(), $source->isConfig);
+		if (!$source->exists) {
+			$source->exists = ($path !== false && is_file($path));
 		}
-		if ($source->exists && $path) {
+		if ($source->exists && $path !== false) {
 			$source->timestamp = filemtime($path);
+		} else {
+			$source->timestamp = 0;
 		}
 	}
 
--- a/src/Template/Compiled.php
+++ b/src/Template/Compiled.php
@@ -248,7 +248,7 @@ class Compiled extends GeneratedPhpFile {
 		if (function_exists('opcache_invalidate')
 			&& (!function_exists('ini_get') || strlen(ini_get("opcache.restrict_api")) < 1)
 		) {
-			opcache_invalidate($this->filepath, true);
+			opcache_invalidate($this->filepath, false);
 		} elseif (function_exists('apc_compile_file')) {
 			apc_compile_file($this->filepath);
 		}
--- a/src/Template/Compiled.php
+++ b/src/Template/Compiled.php
@@ -241,7 +241,7 @@ class Compiled extends GeneratedPhpFile {
 	 * HHVM requires a workaround because of a PHP incompatibility
 	 *
 	 * @param Template $_smarty_tpl do not change/remove variable name, is used by compiled template
-	 * @param bool $invalidateCachedFiles
+	 * @param bool $invalidateCachedFiles forces a revalidation of the file in opcache or apc cache (if available)
 	 *
 	 */
 	private function loadCompiledTemplate(Template $_smarty_tpl, bool $invalidateCachedFiles = true) {
--- a/src/Template/Compiled.php
+++ b/src/Template/Compiled.php
@@ -246,14 +246,14 @@ class Compiled extends GeneratedPhpFile {
 	 */
 	private function loadCompiledTemplate(Template $_smarty_tpl, bool $invalidateCachedFiles = true) {
         
-        if ($invalidateCachedFiles) {
-            if (function_exists('opcache_invalidate')
-                && (!function_exists('ini_get') || strlen(ini_get("opcache.restrict_api")) < 1)
-            ) {
-                opcache_invalidate($this->filepath, false);
-            } elseif (function_exists('apc_compile_file')) {
-                apc_compile_file($this->filepath);
-            }
+		if ($invalidateCachedFiles) {
+			if (function_exists('opcache_invalidate')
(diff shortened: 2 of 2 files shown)


wisskid:
I really wonder why this line was in here in the first place. I guess it was already present in Smarty 4, which makes me wonder if this actually fixes #1007. If so, this must mean there some unintended change in the flow between v4 and v5.

stephanlueckl:
@wisskid 
You are right, this fixes only the symptom of raising wasted memory.
The Smarty5 method `\Smarty\Template\Compiled::loadCompiledTemplate` is the same like in Smarty4

The root cause is in the different call mechanism between Smarty 4 and Smarty 5

**Smarty 4**  loadCompiledTemplate is only called when it was new compiled
```php
public function process(Smarty_Internal_Template $_smarty_tpl)
    {
        $source = &$_smarty_tpl->source;
        $smarty = &$_smarty_tpl->smarty;
        if ($source->handler->recompiled) {
            $source->handler->process($_smarty_tpl);
        } elseif (!$source->handler->uncompiled) {
            if (!$this->exists || $smarty->force_compile
                || ($_smarty_tpl->compile_check && $source->getTimeStamp() > $this->getTimeStamp())
            ) {
                $this->compileTemplateSource($_smarty_tpl);
                $compileCheck = $_smarty_tpl->compile_check;
                $_smarty_tpl->compile_check = Smarty::COMPILECHECK_OFF;
                $this->loadCompiledTemplate($_smarty_tpl);
                $_smarty_tpl->compile_check = $compileCheck;
            } else {
// Loading an already compiled file (standard case)
                $_smarty_tpl->mustCompile = true;
                @include $this->filepath;
// after the include statement mustcompile is false again 
                if ($_smarty_tpl->mustCompile) {
                    $this->compileTemplateSource($_smarty_tpl);
                    $compileCheck = $_smarty_tpl->compile_check;
                    $_smarty_tpl->compile_check = Smarty::COMPILECHECK_OFF;
                    $this->loadCompiledTemplate($_smarty_tpl);
                    $_smarty_tpl->compile_check = $compileCheck;
                }
            }
            $_smarty_tpl->_subTemplateRegister();
            $this->processed = true;
        }
    }
```

**Smarty 5** loadCompiledTemplate is always used, 
```php
private function compileAndLoad(Template $_smarty_tpl) 
	{
		if ($_smarty_tpl->getSource()->handler->recompiled) {
			$this->recompile($_smarty_tpl);
			return;
		}

		if ($this->exists && !$_smarty_tpl->getSmarty()->force_compile
			&& !($_smarty_tpl->compile_check && $_smarty_tpl->getSource()->getTimeStamp() > $this->getTimeStamp())
		) {
// Loading an already compiled file (standard case)		
			$this->loadCompiledTemplate($_smarty_tpl);
		}

		if (!$this->isValid) {
			$this->compileAndWrite($_smarty_tpl);
			$this->loadCompiledTemplate($_smarty_tpl);
		}

		$this->processed = true;
	}
```

Thinking of invalidation is only necessary, when the file is changed in the process.
I would suggest an optional invalidation param to the loadCompiledTemplate() method.
Change will soon follow in the pull request.

As of the forced invalidation, i still think this is not necessary - if the file hasn't change, why force to reload it?

stephanlueckl:
Fix is working for this issue, Fast growing of the wasted memory of OPCache stopped. 

image
Update on this server went online at 16:05.
No more OPCache restarts, CPU spikes stopped.

wisskid:
Nice work @stephanlueckl !

wisskid (approved):

wisskid:
> As of the forced invalidation, i still think this is not necessary - if the file hasn't change, why force to reload it?

I honestly don't have a clue, but there may be some reason both of us don't understand yet, so I'd be more comfortable to leave it in. Your PR looks great, I'm running the tests now.

stephanlueckl:
> I honestly don't have a clue, but there may be some reason both of us don't understand yet, so I'd be more comfortable to leave it in. Your PR looks great, I'm r

Okay, i've changed this line back in the last commit

tungps250991:
@stephanlueckl 
Thank you for the work, it really saves us!

@wisskid
I tried applying this solution & the performance issue [here](https://github.com/smarty-php/smarty/issues/1007#issuecomment-2250038642) is gone. 
Could you please merge this to new v.5 release soon? It would be very helpful since we're going to release the product with Smarty upgrade by the end of this week :)
Thank you!

wisskid:
Expect a new release around Aug 15th.

src/Resource/FilePlugin.php
```diff
@@ -56,6 +56,7 @@ public function populate(Source $source, Template $_template = null) {
 	 * @param Source $source source object
 	 */
 	public function populateTimestamp(Source $source) {
+		$path = false;
```
wisskid, line 59: How about this?
```
public function populateTimestamp(Source $source) {
		$path = $this->getFilePath($source->name, $source->getSmarty(), $source->isConfig);

		if (!$source->exists) {
			$source->timestamp = $source->exists = ($path !== false && is_file($path));
		}
		if ($source->exists) {
			$source->timestamp = ($path !== false && filemtime($path));
		}
	}
```

stephanlueckl: Yes, i see,
Thinking of timestamp is only int, would not reset it to (bool(true)==1) 
'($path !== false && filemtime($path)' is always bool, changed it to an int value.
And explicit reset to 0 (or maybe is null better) if absent

```
	public function populateTimestamp(Source $source) {
              $path = $this->getFilePath($source->name, $source->getSmarty(), $source->isConfig);
              if (!$source->exists) {
                  $source->exists = ($path !== false && is_file($path));
              }
              if ($source->exists && $path !== false) {
                  $source->timestamp = filemtime($path);
              } else {
                  $source->timestamp = 0;
              }
	}
```

wisskid: Yes, you are right. Can you add this to your PR?

stephanlueckl: done

PR #1049, opened 2024-07-31
Allow built-in modifiers override

### LINKED ISSUE #1048: Unable to override built-in modifiers.
Hello,

As it could be done in previous versions, with V5 we are not able to override a built-in modifier using registerPlugin function (Even with unregisterPlugin before).

For example :

```
//Will not work, default one in DefaultExtension->getModifierCallback() will be used instead.
$this->unregisterPlugin(Smarty::PLUGIN_MODIFIER,"date_format"); //With or without this line, same result
$this->registerPlugin(Smarty::PLUGIN_MODIFIER,"date_format",array($this,"myDateFormatModifier")); 

//Will not work, defaut one on DefaultExtension->getModifierCompiler() will be used instead (during template compilation)
$this->registerPlugin(Smarty::PLUGIN_MODIFIER,"json_encode",array($this,"myJsonEncode")); 

//Will not work, default one will be used instead (and triggers an error directly with a message to use "join" instead, changing params orders)
$this->registerPlugin(Smarty::PLUGIN_MODIFIER,"implode",array($this,"myImplodeModifier")); 

...

//Will work, as there is no "myownmodifer" defined in DefaultExtension.
$this->registerPlugin(Smarty::PLUGIN_MODIFIER,"myownmodifier",array($this,"myOwnModifier")); 
```

By Searching in the issues, i have seen that there is a workaround, as it's suggested in #1011 (kind of describing that problem with a custom "json_encode" modifier).
Solution seems to be to create a custom Extension, implementing getModifierCompiler() and getModifierCallback() functions, and register it between Core and Defaut extensions. In that way our Custom Extension car handle modifiers before DefautExtension does it.

But it's not that easy for all cases, as it brings complexity with the different types of modifiers :
"Compiler" Ones (if i understand correctly, executed first when creating compiled templates files), and "Classic" Ones (executed after).

For example, let's say that i want to handle a modifier in MyCustomExtension->getModifierCallback() (not at compilation), if that modifier exists in DefautExtension->getModifierCompiler(), my implementation will be (silently) ignored too. That's because DefaultExtension->getModifierCompiler() will be executed way before MyCustomExtension->getModifierCallback();

So to summarize, if i understand correctly, and from what i've tested so far, for now with v5, every time i want to add/handle a modifier, i have to take care of these points :

- If it's not handled in DefaultExtension : 
Best scenario, i can do whatever i want.
I can use registerPlugin function.
Or i can handle it in a "MyCustomExtension" extension.

- If it's handled in DefaultExtension->getModifierCallback() :
I can't use registerPlugin function (it will be silently ignored)
I can handle it in  MyCustomExtension->getModifierCallback() or ->getModifierCompiler(), as both will be executed before DefaultExtension->getModifierCallback().

- If it's handled in DefaultExtension->getModifierCompiler() :
I can't use registerPlugin function (it will be silently ignored)
I can't handle it in MyCustomExtension->getModifierCallback() (it will be silently ignored too)
I have to handle the modifier in MyCustomExtension->getModifierCompiler() (i have no other choice).

- And in the future updates of Smarty, if DefaultExtension is changed (adding more modifiers, or moving from a type to another), then MyCustomExtension could be (silently) obsolete.

I'm not very enthousiastic with that Extension solution, it's not very intuitive or user fiendly, and seems to required to think about more cases than it appears.
It would be very very much simpler if by default, adding a modifier will take the lead over default ones defined in DefaultExtension.

**As soon as we use "registerPlugin" function, don't we exepect that Smarty use the callback we just explicitly defined ?**

2Vince:
A proposition to fix #1048 (and may be a part of #1011 too)

This is to prevent modifiers added by registerPlugin function from beeing silently ignored by DefaultExtension.
More details in #1048

### DIFF
--- a/src/Extension/DefaultExtension.php
+++ b/src/Extension/DefaultExtension.php
@@ -6,13 +6,26 @@ use Smarty\Exception;
 
 class DefaultExtension extends Base {
 
+    /**
+     * @var \Smarty\Smarty
+     */
+    private $smarty;
+
 	private $modifiers = [];
 
 	private $functionHandlers = [];
 
 	private $blockHandlers = [];
 
+    public function __construct(\Smarty\Smarty $smarty) {
+        $this->smarty = $smarty;
+    }
+
 	public function getModifierCompiler(string $modifier): ?\Smarty\Compile\Modifier\ModifierCompilerInterface {
+
+        if($this->smarty->getRegisteredPlugin(\Smarty\Smarty::PLUGIN_MODIFIER, $modifier) ?? $this->smarty->getRegisteredPlugin(\Smarty\Smarty::PLUGIN_MODIFIERCOMPILER, $modifier)) {
+            return null;
+        }
 
 		if (isset($this->modifiers[$modifier])) {
 			return $this->modifiers[$modifier];
@@ -53,7 +66,12 @@ class DefaultExtension extends Base {
 	}
 
 	public function getModifierCallback(string $modifierName) {
-		switch ($modifierName) {
+
+        if ($this->smarty->getRegisteredPlugin(\Smarty\Smarty::PLUGIN_MODIFIER, $modifierName)){
+            return null;
+        }
+
+        switch ($modifierName) {
 			case 'capitalize': return [$this, 'smarty_modifier_capitalize'];
 			case 'count': return [$this, 'smarty_modifier_count'];
 			case 'date_format': return [$this, 'smarty_modifier_date_format'];
--- a/src/Smarty.php
+++ b/src/Smarty.php
@@ -526,7 +526,7 @@ class Smarty extends \Smarty\TemplateBase {
 		$this->BCPluginsAdapter = new BCPluginsAdapter($this);
 
 		$this->extensions[] = new CoreExtension();
-		$this->extensions[] = new DefaultExtension();
+		$this->extensions[] = new DefaultExtension($this);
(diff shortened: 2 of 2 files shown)


wisskid:
Sorry, but I don't like this approach. The registerplugin methods are the "old" way to register plugins. I'm in favor of creating new Extensions.

Wouldn't it be easier to switch the default loading order of the Extensions? It is now: core, default, BC. We could change it to core, BC, default.

wisskid:
> I understand you consider "registerPlugin" method as old way to register Plugins. Does this mean that it could be deprecated in a near future ?

Yes, it might.

wisskid:
That part didn't really change, but you are right, that would be very useful. Feel free to drop a MR if you can find the time for it.

hron84:
You misunderstood me. I cannot figure this out on my own based on the current docs and I am not THAT expert in PHP/Smarty internals to figure out on my own... 😞

I'm trying to upgrade a codebase and I am basically stuck with it without knowing how things work... 😅

wisskid:
@hron84 I agree that part is not too well documented. However, if you browser the Smarty codebase, there are tons of examples to draw from. For example, open https://github.com/smarty-php/smarty/blob/master/src/Extension/DefaultExtension.php and then click on to the different ModifierCompilers, ModifierCallbacks, FunctionHandler, et cetera to get a feel for how they work.

PR #1067, opened 2024-09-18
Smarty 4 Suppress Deprecation Notices for Unregistered Callables in Templates using muteUndefinedOrNullWarnings function

dipeshsukhia:
Hi @wisskid

As per discussion [#967](https://github.com/smarty-php/smarty/discussions/967), Smarty 4.5 has introduced deprecation notices to warn users about the usage of unregistered callables in templates. 

To prevent these warnings from cluttering the logs, we've implemented a fix that suppresses these deprecation notices by utilizing the muteUndefinedOrNullWarnings function for Smarty 4.5.

This change ensures that deprecated notices for unregistered callables are properly handled, preventing unnecessary interruptions 
 in Smarty 4.5.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [4.5.5] - 2024-09-18
+- smarty 4 suppresses Smarty register class-plugin error in muteUndefinedOrNullWarnings function
+
 ## [4.5.4] - 2024-08-14
 - Fixed that using `count()` would trigger a deprecation notice. [#813](https://github.com/smarty-php/smarty/issues/813)
 
--- a/libs/sysplugins/smarty_internal_errorhandler.php
+++ b/libs/sysplugins/smarty_internal_errorhandler.php
@@ -30,6 +30,18 @@ class Smarty_Internal_ErrorHandler
     public $allowUndefinedArrayKeys = true;
 
     /**
+     * Allows Unregistered smarty Plugin.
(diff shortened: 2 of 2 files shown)


wisskid:
Why would Smarty throw a notice first and then catch and supress it again? The notices are there for a reason.

If you don't want the depreciation notices, just change your error logging (or reporting) levels.

dipeshsukhia:
Hi @wisskid

Thank you for your feedback!

The goal of suppressing deprecation notices at the package level using muteUndefinedOrNullWarnings is to give developers flexibility without having to modify global error logging settings. This approach is especially helpful for those working with legacy code or third-party integrations, allowing them to maintain clean logs without missing critical issues.

This solution specifically targets unregistered callables while preserving the ability to manage other important notices.

wisskid:
I think I see what you mean, but am not convinced that your approach is correct.

First of all: `muteUndefinedOrNullWarnings` has a very different goal. It mutes warnings (not notices) and does so in order to maintain identical behavior when switching from PHP7 to PHP8. Nobody should have any reason to expect that calling this method would drop the newly added deprecation notices in Smarty. If there was an important use case that calls for preventing these notices from ending up in logs, there should be a (separate) setting that would prevent the notices from being thrown in the first place.

However, that goal "to maintain clean logs without missing critical issues" seems quite arbitrary. You seem to feel that the notices are not critical and other notices are more important, but other people might disagree or have additional preferences.

it seems to me that you could just as easily add a custom error handler to the application that is using Smarty, temporarily change our error logging settings when calling `$smarty->display('template.tpl')` or configure your logging output stream to drop certain messages if this is something you need.

PR #1043, opened 2024-07-12
Add PHP 8.4 support to Smarty

Visualq:
This pull request updates the code base to support the deprecations introduced in PHP 8.4. It should be backwards compatible.

1. Updated the function signatures that had implicit null assignment to typed parameters. 
2. Added PHP 8.4 to the 'run tests'-shell script.
3. Bumped the PHP version in the README.md

PR #1084, opened 2024-11-21
Add PHP 8.4 support to Smarty

jolelievre:
This PR is mostly a backport of https://github.com/smarty-php/smarty/pull/1043 for version 4

This pull request updates the code base to support the deprecations introduced in PHP 8.4. It should be backwards compatible.

Updated the function signatures that had implicit null assignment to typed parameters.
Added PHP 8.4 to the 'run tests'-shell script.
Bumped the PHP version in the README.md

### DIFF
--- a/libs/sysplugins/smartycompilerexception.php
+++ b/libs/sysplugins/smartycompilerexception.php
@@ -21,7 +21,7 @@ class SmartyCompilerException extends SmartyException
         int $code = 0,
         ?string $filename = null,
         ?int $line = null,
-        Throwable $previous = null
+        ?Throwable $previous = null
     ) {
         parent::__construct($message, $code, $previous);
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,7 @@ jobs:
           - "8.1"
           - "8.2"
           - "8.3"
+          - "8.4"
 
         compiler:
           - default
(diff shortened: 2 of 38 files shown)


libs/sysplugins/smartycompilerexception.php
```diff
@@ -14,14 +14,14 @@ class SmartyCompilerException extends SmartyException
...
-     * @param Throwable|null $previous The previous exception used for the exception chaining.
+     * @param \Throwable|null $previous The previous exception used for the exception chaining.
      */
     public function __construct(
         string $message = "",
         int $code = 0,
         ?string $filename = null,
         ?int $line = null,
-        Throwable $previous = null
+        ?\Throwable $previous = null
```
wisskid, line 24: why did you add the backslash?

wisskid:
@jolelievre couple of issues with the unit tests, I'm afraid.

libs/sysplugins/smartycompilerexception.php
jolelievre: Mostly because it was also added in the PR for V5 https://github.com/smarty-php/smarty/pull/1043/files#diff-d0f40f666506e3868faaf237c1f5a7731be2cd3c96bfa986b564f7c765b42bb5R26

But I can remove it

jolelievre:
Ok thanks @wisskid I'll look into it

libs/sysplugins/smartycompilerexception.php
jolelievre: Ah but I'm dumb there's no namespace in smarty 4, so not needed

jolelievre:
@wisskid it should be good now, at least locally I don't have any errors anymore, I had missed some E_STRICT
And I remove the `\` for the `Throwable`

wisskid (approved):

jolelievre:
Thanks @wisskid

Issue #1100, opened 2025-01-29
Modifier called like a function compiled to modifier name, not function callback


rudiedirkx:
Smarty 4.5.5

Is this a bug or a very strange feature? If I define a modifier:

```php
$this->smarty->registerPlugin('modifier', 'kprint_r_out', trim(...));
```

and call it like:

```smarty
{kprint_r_out($myVar)}
```

it will compile into

```php
<?php echo kprint_r_out($_smarty_tpl->tpl_vars['myVar']->value);?>
```

even though `kprint_r_out` is only the **name** of the function, not the function itself (but `kprint_r_out()` does exist, otherwise it won't compile). The function (`trim` in this case) isn't used at all...

I like calling functions like function, but this is too weird. If it's intentional, it shouldn't be defined as a modifier with mandatory callback (that's not used).

Or am I abusing another Smarty functionality somehow?

rudiedirkx:
I also do like some functions being compiled into native functions (instead of going through the plugin system), but if I explicitly define a different callback, it should probably use that, not the function name.

wisskid:
Could you provide a full code example? From your report I have several questions.

1. `trim(...)` would return a string, which would be the callback registered by smarty
2. You seem to call `dump` instead of `kprint_r_out` in your template. I don't see `dump` being registered.

rudiedirkx:
All kinds of callbacks, and how to reflect their source: https://3v4l.org/D8hib

I should even be able to make a Smarty modifier like:

```php
$manager = getAccessManager();
$this->smarty->registerPlugin('modifier', 'can', $manager->hasAccess(...));
```

And then call `{if can($arg)}` without object context, which calls `$manager->hasAccess($arg)`. Valid PHP, and Smarty doesn't complain when registering it, only it's not compiled correctly.

rudiedirkx:
> only it's not compiled correctly

It might be compiled correctly for use as actual modifier though: `{$myVar|can}`. That might have a different compiler. That's why I ask. It seems odd that you can call a modifier like a real function. Is that intentional?

rudiedirkx:
Super simple buggy way to use Reflection to compile those callbacks into code: https://3v4l.org/m0tP7 Sometimes it's possible to compile into native callable, but sometimes it has to go through the plugin systems.

wisskid:
Aaah got it now! I'll look into it.

wisskid:
I wrongfully assumed the three dots meant you left some irrelevant code out... 😣

[...]

PR #1101, opened 2025-02-03
Bugfix/1100 modifier called like a function compiled to modifier name not callback

wisskid:


wisskid:
Fixes #1100

rudiedirkx:
Yes, this fixes. But now every call goes through the plugin system:

```
<?php if (call_user_func_array($_smarty_tpl->registered_plugins[ 'modifier' ][ 'can' ][ 0 ], array( 'foo' ))) {?>
        <?php echo call_user_func_array($_smarty_tpl->registered_plugins[ 'modifier' ][ 'dump' ][ 0 ], array( $_smarty_tpl->tpl_vars['title']->value ));?>
<?php }?>
```

and I did like the efficiency of the previous (bug buggy) compilation:

```
<?php if (can('foo')) {?>
```

but this is definitely better. Bug fixed. 👍

wisskid:
Actually, I don't think that changed (it was compiled like that anyway), but when investigating, I found unintentional changes which I fixed in the last commit.

Issue #1119, opened 2025-03-28
PHP 8.4 support


santaclaus21:
In this pull request (https://github.com/smarty-php/smarty/pull/1043/files), you have omitted /src/Resource/StringEval.php:34.
It produces the following warning on PHP 8.4:
Deprecated: Smarty\Resource\StringEval::populate(): Implicitly marking parameter $_template as nullable is deprecated; the explicit nullable type must be used.

gherosh:
Any updates on this?

wisskid:
@gherosh thanx for the PR!

alexgit2k:
@wisskid Even though a new version was released just 2 day ago, could you please release a new version? It looks like this is the last deprecation notice for PHP 8.4.

wisskid:
@alexgit2k sure, done!

PR #1126, opened 2025-04-14
Update StringEval.php

gherosh:
Fix: Implicitly marking parameter $_template as nullable is deprecated, the explicit nullable type must be used instead


gherosh:
Fixes #1119

PR #1130, opened 2025-05-04
Add PHPUnit tests for Smarty scopes and fix template source handling

LaKanDoR:
- Introduced comprehensive PHPUnit tests for various Smarty scope functionalities, including append and assign scopes.
- Implemented tests for caching behavior and template inclusion with different scopes.
- Added error handling for undefined template variables and improved test coverage for caching mechanisms.
- Updated test results files to reflect the latest test outcomes and deprecations.
- Fixed issues related to argument counts in cache resource tests.
- Created missing template files for testing purposes.


wisskid:
@LaKanDoR this is a massive PR, adding a lot of (seemingly unnecessary) files. I'm a big fan of more unit tests, especially for scope matters, but I also notice that this would drop support for a bunch of PHP versions (which we don't want). What are you trying to do here? What problem are you solving?

Issue #1005, opened 2024-04-26
Shorthand Syntax


SleepingForge:
So in `v5.0.0-rc3` I could register a Plugin-Function and do something like this:

`{myFunc test1 arg='arg'}`

because of this fix: #953 it's not possible anymore to assign string attributes, because of this section in Compiler/Base.php:

```php
if (!is_array($mixed)) {
			// options flag ?
			if (isset($options[trim($mixed, '\'"')])) {
				$_indexed_attr[trim($mixed, '\'"')] = true;
				// shorthand attribute ?
			} elseif (isset($this->shorttag_order[$key])) {
				$_indexed_attr[$this->shorttag_order[$key]] = $mixed;
			} else {
				// too many shorthands
				$compiler->trigger_template_error('too many shorthand attributes', null, true);
			}
			// named attribute
		}
```

Because shorttag_order wasn't empty it would just insert Attribute `test1` into `$_indexed_attr` and I could use it as ['var1'].
I found it quite useful cause the first argument could be used as some kind of identifier.

I found old issues that discuss this shorthand syntax for example: #165, #342, #568

So is this Syntax officially dead in Smarty v5.0.0? Or is there a way to kind of inject the first parameter still as a string?

Thank you!

wisskid:
The var1, var2, var3 shorttag order got into the release candidates for smarty 5 by accident. They didn't exist in Smarty v4 either.

We should be able to support this in a future version, but I think you should have a way to specify the number and names of the parameters. We might be able to extend FunctionHandlerInterface for this, which would allow you to write an Extension that supports shorthand attributes.

PR #1142, opened 2025-09-09
Add support for shorttags in functions

pharixces:
This fixes https://github.com/smarty-php/smarty/issues/1005

### DIFF
--- a/src/Compile/FunctionCallCompiler.php
+++ b/src/Compile/FunctionCallCompiler.php
@@ -3,39 +3,17 @@
  * Smarty Internal Plugin Compile Registered Function
  * Compiles code for the execution of a registered function
  *
-
-
  * @author     Uwe Tews
  */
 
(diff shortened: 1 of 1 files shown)


pharixces:
Forgot to run the unittests, will look into it.

pharixces:
Added an extra interface `AttributeFunctionHandlerInterface` which extends the `FunctionHandlerInterface`. The functioncall compiler checks whether the functionhandler implements this interface and sets the attributes of the functioncall compiler. This way you can add your own shorttags (and other attributes) to your functionhandler. 

I'm not quite satisfied with the unset($_attr['nocache']); in the functioncallcompiler class. It almost seems this is only added to make the unittests complete for the assign function within smarty. Gonna take a closer look at this.

wisskid (changes requested):

wisskid (approved):

PR #1164, opened 2026-01-07
Fix static analysis warnings for isDot() and remove deprecated APC support

### LINKED ISSUE #1144: Errors in Smarty.php version 5.5
line 1347 : `if (!$_compile->isDot()) {`
Undefined method 'isDot'.

line 1389 : apc_delete_file($_filepath);
Undefined function 'Smarty\apc_delete_file'.

h-marumoto:
This PR fixes #1144

This issue likely refers to errors generated by static analysis.
I have addressed those issues.

Undefined method 'isDot'.

#### Investigation
While investigating this warning, I discovered that:
1. isDot() is not available on the loop variable
RecursiveIteratorIterator returns SplFileInfo objects in foreach loops
SplFileInfo does not have an isDot() method
isDot() only exists on DirectoryIterator and RecursiveDirectoryIterator classes
2. The isDot() check was unnecessary
All entries starting with . are already filtered out earlier:
Since isDot() only matches . and .. (both start with .), they are already excluded
The isDot() check is unreachable code
```
if (substr(basename($_file->getPathname()), 0, 1) === '.') {
    continue;
}
```
#### Fix
Removed the unnecessary isDot() checks from:
- src/Smarty.php (line 1347)
- src/Cacheresource/File.php (line 224)
- tests/PHPUnit_Smarty.php (line 294)
#### Impact
No functional changes.
The removed checks were redundant and never executed.

Undefined function 'Smarty\apc_delete_file'.

The old APC (Alternative PHP Cache) extension was deprecated in PHP 5.5 and is not available in PHP 7.0+.
Since Smarty now requires PHP 7.2+ (as defined in composer.json), all APC-related code has become dead code that will never execute.
#### Fix:
- Removed dead code in Smarty.php and src/Cacheresource/File.php
- Removed APC cache resource implementation and tests:
  - tests/UnitTests/__shared/cacheresources/cacheresource.apc.php
  - tests/UnitTests/CacheResourceTests/_shared/PHPunitplugins/cacheresource.apctest.php
  - tests/UnitTests/CacheResourceTests/Apc/CacheResourceCustomApcTest.php
#### Impact:
No functional impact.
The removed code was never executed on supported PHP versions (7.2+).

I would appreciate it if you could review this.
Thank you!

### DIFF
--- a/src/Cacheresource/File.php
+++ b/src/Cacheresource/File.php
@@ -221,10 +221,8 @@ class File extends Base
 			    $_filepath = (string)$_file;
 			    // directory ?
 			    if ($_file->isDir()) {
-				    if (!$_file->isDot()) {
-					    // delete folder if empty
-					    @rmdir($_file->getPathname());
-				    }
+                    // delete folder if empty
+                    @rmdir($_file->getPathname());
 			    } else {
 				    // delete only php files
 				    if (substr($_filepath, -4) !== '.php') {
--- a/src/Smarty.php
+++ b/src/Smarty.php
@@ -1344,10 +1344,8 @@ class Smarty extends \Smarty\TemplateBase {
 			}
 			$_filepath = (string)$_file;
 			if ($_file->isDir()) {
-				if (!$_file->isDot()) {
-					// delete folder if empty
-					@rmdir($_file->getPathname());
-				}
+				// delete folder if empty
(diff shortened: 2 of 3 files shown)


wisskid:
good finds! And if the `isDot` method isn't available, wouldn't the entire `isDir` if-branch surrounding it be useless as well?

PR #1166, opened 2026-01-14
Fix static analysis errors in Data class method chaining

### LINKED ISSUE #1143: Error chaining assign and display
VSC signals an error Undefined method 'display' when chaining assign and display
`(new Smarty())->assign("x", "x")->display('template.tpl');`

But it works

h-marumoto:
This PR fixes #1143

This PR resolves static analysis errors when using method chaining.

The `Smarty` class extends the `Data` class, but the PHPDoc return types for methods in the `Data` class (`assign()`, `append()`, etc.) were documented as returning `Data`.

As a result, when using method chaining like the following, static analysis tools would interpret that a `Data` instance is returned instead of a `Smarty` instance, and would report errors that `Smarty`-specific methods (e.g., `display()`) do not exist.

```php
(new Smarty())->assign("x", "x")->display('template.tpl');
// Undefined method 'display'.
```

I changed the PHPDoc return type from Data to $this. By using $this, static analysis tools can correctly recognize that an instance of the actually invoked class (in this case, Smarty) is returned.

This change only modifies PHPDoc comments and has no effect on runtime behavior.

### DIFF
--- a/src/Data.php
+++ b/src/Data.php
@@ -100,7 +100,7 @@ class Data
 	 * @param boolean $nocache if true any output of this variable will be not cached
 	 * @param int $scope one of self::SCOPE_* constants
 	 *
-	 * @return Data current Data (or Smarty or \Smarty\Template) instance for
+	 * @return $this current Data (or Smarty or \Smarty\Template) instance for
 	 *                              chaining
 	 */
     public function assign($tpl_var, $value = null, $nocache = false, $scope = null)
@@ -162,7 +162,7 @@ class Data
      * @param bool         $nocache if true any output of this variable will
      *                              be not cached
      *
-     * @return Data
+     * @return $this
      * @api  Smarty::append()
      */
 	public function append($tpl_var, $value = null, $merge = false, $nocache = false)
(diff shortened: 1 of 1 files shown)


wisskid:
Hi @h-marumoto I see how this fix would work and would probably be fine for most cases, but `@return $this` doesn't seem to be official phpdoc syntax?  In your defense: the codebase already contains 3 instances of `@return $this`  in the `Smarty\Smarty` class.

h-marumoto:
Hi @wisskid , sorry for the slow response!

Fair point — strictly speaking `@return $this` was never ratified.
It is, however, officially supported by phpDocumentor, and PHPStan, Psalm, and PhpStorm all recognize it — PHPStan even lists `$this` as a basic type in its PHPDoc parser, and Psalm internally treats it as a type token equivalent to `static`.
Ref: https://docs.phpdoc.org/guide/references/phpdoc/tags/return.html#the-this-keyword

I'd prefer to keep `$this` here since it's the most precise annotation for these methods (the same instance is returned, not just an instance of the called class), and it matches the existing usage in `Smarty\Smarty`.
Let me know if you'd like any changes!

PR #1167, opened 2026-01-19
Add type validation for security_policy property.

reomin:
## Context
This PR fixes some issues where SecurityPolicy checks are silently skipped when the internal security_policy property is overwritten with a non-object value.

Currently, several SecurityPolicy enforcement paths rely on isset() and is_object() checks. If security_policy is set but is not an object, these checks are bypassed without any error, even when enableSecurity() has been called.

This PR makes the behavior more robust and consistent with the intended design of SecurityPolicy by enforcing stricter type handling.

## Changes

- Changed `security_policy` from `public` to `private`
- Added `getSecurityPolicy()` and `setSecurityPolicy()` methods
- Added `__set()` method to validate type on assignment
- Updated all internal code to use getter/setter instead of direct property access
- Updated tests to use `getSecurityPolicy()`

### DIFF
--- a/src/Smarty.php
+++ b/src/Smarty.php
@@ -246,9 +246,9 @@ class Smarty extends \Smarty\TemplateBase {
 	/**
 	 * implementation of security class
 	 *
-	 * @var \Smarty\Security
+	 * @var \Smarty\Security|null
 	 */
-	public $security_policy = null;
+	private $security_policy = null;
 
 	/**
 	 * debug mode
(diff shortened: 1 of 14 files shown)


src/Smarty.php
```diff
@@ -2231,5 +2252,28 @@ public function setCacheModifiedCheck($cache_modified_check): void {
...
+			if ($value !== null && !($value instanceof \Smarty\Security)) {
+				$given = is_object($value) ? get_class($value) : gettype($value);
+				throw new \Smarty\Exception(
+					'security_policy must be null or \Smarty\Security, ' . $given . ' given'
+				);
+			}
+			$this->setSecurityPolicy($value);
+			return;
+		}
+		$this->$name = $value;
```
wisskid, line 2275: I'm worried about this line. Is adding a magic setter method to the Smarty class safe enough? there used to be a lot of direct property setting in smarty before version 4.

reomin:
HI, @wisskid 

Thank you for your feedback!
That makes sense. I hadn't fully considered the impact of `__set()` on code written before version 4. 
Given that, I don't think there is a clean way to fix this without introducing other risks. 
I will close this PR for now.

PR #1182, opened 2026-04-24
Fix: incorrect `strpos` check in `triggerTagNoCache `

zigzagdev:
## Description                                                                                                                                                                                   
                                                                                                                                                                                                
  ### Summary                                                                                                                                                                                    
  `triggerTagNoCache()` uses `!strpos($variable, '(')` to check if `(` is absent                                                                                                                
  from `$variable`. However, `strpos` returns `0` when `(` appears at position 0,                                                                                                               
  and `!0` evaluates to `true`, incorrectly skipping the nocache flag logic.                                                                                                                    
                                                                                                                                                                                                
  This replaces `!strpos(...)` with `strpos(...) === false`.                                                                                                                                    
                                                                                                                                                                                                
  ## References                                                                                                                                                                                 
  - https://www.php.net/manual/en/function.strpos.php#refsect1-function.strpos-notes

### DIFF
--- a/src/Compiler/Template.php
+++ b/src/Compiler/Template.php
@@ -506,7 +506,7 @@ class Template extends BaseCompiler {
 	 * @return string
 	 */
 	public function triggerTagNoCache($variable): void {
-		if (!strpos($variable, '(')) {
+		if (strpos($variable, '(') === false) {
 			// not a variable variable
 			$var = trim($variable, '\'');
 			$this->tag_nocache = $this->tag_nocache |


wisskid:
Would this ever be called when when `(` is at
position 0?

zigzagdev:
@wisskid

Exactly ! You're right — in current usage, `(` might not appear at position 0.

But there are two separate concerns:

1. **Current input patterns**: May not trigger in this pattern.
2. **Code correctness**: `strpos()` returns 0 at position 0, and !0 evaluates to true which is logically incorrect regardless of current inputs

Even if position 0 doesn't occur now. Because there are three reasons

1. It's unsafe against the function's contract
2. It aligns with PHP best practice (PHP docs explicitly `recommend === false`)
3. It protects against future refactoring or input format changes

This is a low-risk correctness fix, not just a style improvement.

PR #1206, opened 2026-08-17
patch propose deprecated php8.6

klaoun2:
hello
i propose this pr


Amaury:
Did you test your code before submitting it? This PR does not parse. The three commits copy the function *signatures* from the php.net manual into the *call sites*:

```php
$mb_reg_charset = mb_regex_encoding(?string $encoding = null):
$parts = mb_split(preg_quote(string $search), string $subject ?? "") ?: array();
```

Parameter types and defaults are only legal in a declaration, plus the statement terminators are `:` instead of `;`.

`php -l src/functions.php` fails, and since Composer autoloads that file via `files`, this is a fatal error on every PHP version.

